PR #4, opened 2016-06-29
Added bootstrap-sass gem and made the relevant changes to application…

chadbaum:
….scss, application.js, and application.html.erb

PR #5, opened 2016-06-29
Update rails to latest stable

chadbaum:
Just updating to latest stable rails, I don't see any reason we should be on a lower version.  I've used it in my personal projects with the gems we are familiar with and have had no issues.


lasley:
This PR is dependent on merge of #4 and is 👎  until that time.

I recommend basing all new branches from master instead of your most recent branch. This will reduce dependencies & conflict possibilities.

chadbaum:
Whoops.  Will do so in future.

aleksgorbenko:
I think it's good to go now, the branch just needs to pull the latest master in order to merge.

lasley:
LGTM

PR #9, opened 2016-07-06
Add piece model and inherit individual pieces

ZachBerman:
STI dictates adding a type column to our pieces table

-All chess pieces (pawn, rook, knight, bishop, queen, king) inherit from piece model
-Piece model validates the above strings are allowed only

### DIFF
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -2,5 +2,5 @@ class Piece < ActiveRecord::Base
   belongs_to :game
   belongs_to :player
 
-  validates :type, inclusion: { in: %w(pawn rook bishop king queen) }
+  validates :type, inclusion: { in: %w(pawn rook bishop knight king queen) }
 end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -3,4 +3,5 @@ class Piece < ActiveRecord::Base
   belongs_to :player
 
   validates :type, inclusion: { in: %w(pawn rook bishop knight king queen) }
+  validates :color, inclusion: { in: %w(black white) }
 end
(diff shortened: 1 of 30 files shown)


app/models/piece.rb
```diff
@@ -0,0 +1,6 @@
+class Piece < ActiveRecord::Base
+  belongs_to :game
+  belongs_to :player
+
+  validates :type, inclusion: { in: %w(pawn rook bishop king queen) }
```
lasley: Missing knight

spec/models/knight_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Knight, type: :model do
```
lasley: You should be testing the creations of all your children. This test would have yielded a failure due to lack of knight in type validation

app/models/piece.rb
```diff
@@ -0,0 +1,6 @@
+class Piece < ActiveRecord::Base
+  belongs_to :game
+  belongs_to :player
+
```
lasley: Maybe a validation for color black/white?

spec/models/piece_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Piece, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
lasley: Add test for failed type validation: `type: no_exist` or similar

ZachBerman:
Ok - so I've fixed a few things but my STI seems to be broken. I can't seem to find a resource that says to do anything differently - any advice on what I am missing?

PR #11, opened 2016-07-09
[WIP] Create piece model, and inherit all individual pieces from it

ZachBerman:
-STI is functioning - tested thoroughly in console but have not been able to create valid rspec tests. Leaving town but will be back Monday to work on this. Wanted to submit this PR so nobody gets held back (believe this task was a blocker).


lasley:
FYI when this is done, make sure to remove the WIP tag and drop a comment letting everyone know it is ready for review again.

PR #12, opened 2016-07-09
Added latest version of sass-rails gem

chadbaum:


aleksgorbenko:
I will merge it as it is basically a fix.

PR #13, opened 2016-07-09
added the view with chessboard on the 'new' page. modified routes.rb

aleksgorbenko:
- Created a view via existing controller (can change later if needed).
- Used Sass mainly for styling.
- Each cell has an ID of A5, D3, etc, so these can be mapped as coordinates later when updating piece position in the database with jQuery (not sure, maybe will have to change to 'data-id' attribute instead).


app/assets/stylesheets/_mixins.scss
```diff
@@ -0,0 +1,11 @@
...
+  box-shadow: inset 0 0 0 $amount $color;
+  -moz-box-shadow: inset 0 0 0 $amount $color;
+  -webkit-box-shadow:inset 0 0 0 $amount $color;
+}
+
+@mixin cell-gradient($from, $to) {
+  background:#cecece;
+  background:-moz-linear-gradient(top, $from, $to);
+  background:-webkit-gradient(linear,0 0, 0 100%, from($from), to($to));
+}
```
lasley: All files should end in a newline. POSIX spec defines a line as ending in a carriage return:

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206
http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

app/assets/stylesheets/application.scss
```diff
@@ -1,2 +1,67 @@
...
+  -webkit-box-shadow:inset 0 0 10px rgba(0,0,0,.4);
+  @include cell-gradient(#ccc, #eee);
+}
+
+/* have to use a class here to override box shadow settings above */
+.container #chess-board tr td:hover {
+  box-shadow: inset 0 0 12px rgba(166,34,25, 0.6);
+  -moz-box-shadow: inset 0 0 12px rgba(166,34,25, 0.6);
+  -webkit-box-shadow: inset 0 0 12px rgba(166,34,25, 0.6);
+}
```
lasley: Newline

app/views/games/_chessboard.html.erb
```diff
@@ -0,0 +1,86 @@
+<div class="container">
+
+
+    <table id="chess-board" class="col-centered" >
+      <tr>
```
lasley: Looks like a lot of repetition here, what about maybe a series of loops to generate the rows and cols dynamically?

app/views/games/_chessboard.html.erb
```diff
@@ -0,0 +1,86 @@
...
+        <td id="B1"></td>
+        <td id="C1"></td>
+        <td id="D1"></td>
+        <td id="E1"></td>
+        <td id="F1"></td>
+        <td id="G1"></td>
+        <td id="H1"></td>
+      </tr>
+    </table>
+</div>
```
lasley: Newline

app/views/games/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<br>
+<br>
+
+<div class="text-center players">
+  <h2>&#9819; Player1 vs &#9813; Player2</h2>
```
lasley: Should these be the names of the users?

lasley:
Reviews inline.

FYI when this is done, make sure to remove the WIP tag and drop a comment letting everyone know it is ready for review again.

app/assets/stylesheets/_mixins.scss
aleksgorbenko: inserted newline - thanks for the links i didnt know it is a convention

app/assets/stylesheets/application.scss
aleksgorbenko: fixed

app/views/games/new.html.erb
aleksgorbenko: as devise is not yet marged - we dont have the user model set, so I just wrote Player 1 & 2 and temporal placeholders

aleksgorbenko:
Chessboard is now created with the loop via `append` jQuery method

lasley:
Hmm that works too, although I was just thinking to do it in ruby. `<% 8.times do |row| %>` `<% 8.times do |even_cell|%>`

Will need one more loop though to generate the cols. General rule of thumb - if you find yourself copy pasting, you are doing it wrong. When developing, you should always be adding, deleting, or CUT pasting. Copy paste is always an indicator of DRY violations.

aleksgorbenko:
You are right, the code was not DRY at all, I just assumed that with HTML (which is not going to be dynamic in the case of chess board) it is ok to write everything explicitly. Will keep in mind that DRY applies _everywhere_ :) 

P.S ever since I discovered cmd+shift+d in Sublime, duplicating anything is a breeze...

General question - is there a matter of efficiency in terms of whether we use Ruby or jQuery to generate HTML dynamically? (we can cover this on a call if that's easier)

lasley:
Basically it's where the processing is happening. Ruby would be server side, jQuery/JS client. 

The downside to rendering client side is that you cannot render Javascript DURING the render of the DOM, so there is a split millisecond where the page will not contain the elements that are being created. 

Additionally even though it looks like you are creating parsed HTML w/ jQuery, you are in fact creating a jQuery object of a bunch of Javascript objects, then appending them to the DOM

```
$('<div><span class='derp'>Test</span></div>').append(document.body);

// Is actually
$( 
  document.createElement('div').appendChild(
    document.createElement('span').appendChild(
      document.createTextNode('Test')
   )
);
```

To contrast this, Ruby would send the parsed HTML directly to the browser to be rendered directly w/out the creation of a Javascript of jQuery object. It ends up being the exact same thing as your explicitly typed example, but allows you to be lazy.

aleksgorbenko:
Ok, @chadbaum I have updated the HTML attributes for each row.

I have also updated the boot.rb file - that shouldn't anything for you guys. Chances are we all have similar issue after updating to the latest Rails 4.2.1 

Now we dont have to run "rails server -b 0.0.0.0" but just "rails server"

chadbaum:
I would consider using Ruby instead of jQuery as @lasley mentioned but otherwise looks good to me.  :+1:

lasley:
I'm good with it like this 👍 - @aleksgorbenko did you want to switch the board creation to ruby, or unify the column creations?

aleksgorbenko:
As I understood, you said @lasley technically there is not much difference between Ruby way and jQuery, so I modified the existing solution.

lasley:
Cool I'm good with that. Resolve the conflicts please, then we'll be good to merge!

aleksgorbenko:
Ok, guys all updated now (had some weird merge conflicts as devise and capybara branch was merged first).  It is ready to merge now.

app/assets/stylesheets/application.scss
chadbaum: I would consider moving these styles to their own SCSS file and import them into application.scss.  My understanding is that is rails best practice instead of actually putting styles in application.css.

app/views/games/new.html.erb
chadbaum: Devise is merged now, so may want to go ahead and update this now.

PR #14, opened 2016-07-09
fixed info about aleks gorbenko

aleksgorbenko:


aleksgorbenko:
only readme file is updated, no other code was edited, so i will merge

PR #16, opened 2016-07-11
Move Validation Logic

chadbaum:
Added validation checks for all moves each piece can make based on the X,Y coordinates of the 2-D game board array.  White player is on bottom of board, Black player is on top (for logic only).  Does not yet include:

- Collision/capture logic for all pieces
- Castling logic (from my estimation, we will need a boolean in the database to track if a rook has ever moved for this one)
- Check logic
- Checkmate logic

I've included about 70 rspec tests for various move combinations with the pieces.  Will expand on it as we add more logic.  I also tried to add comments for all methods where it made sense.  This was built off of zach's PR since I needed his models in place to do this.

EDIT:  Added WIP tag while I process @lasley comments.

### DIFF
--- a/app/models/bishop.rb
+++ b/app/models/bishop.rb
@@ -1,2 +1,7 @@
 class Bishop < Piece
+
+  def valid_move?(x, y) #NO CAPTURE/COLLISION LOGIC YET
+    moved?(x,y) && diagonal_move?(x,y) #bishop moved any number of diagonal spaces
+  end
+
 end
--- a/app/models/bishop.rb
+++ b/app/models/bishop.rb
@@ -1,7 +1,9 @@
 class Bishop < Piece
 
-  def valid_move?(x, y) #NO CAPTURE/COLLISION LOGIC YET
-    moved?(x,y) && diagonal_move?(x,y) #bishop moved any number of diagonal spaces
+  # Capture, collision, check, and checkmate logic are not
+  # implemented yet and thus ignored.
+  def valid_move?(x, y)
+    moved?(x,y) && diagonal_move?(x,y)
   end
 
 end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -7,38 +7,44 @@ class Piece < ActiveRecord::Base
 
   private
 
-  def x_distance(x) #how many tiles piece has moved on x-axis
+  #ALL MOVE CHECKS ASSUME WHITE PLAYER IS AT BOTTOM OF 2-D ARRAY/BOARD, AND BLACK PLAYER IS ON TOP.  NO IMPACT ON VIEW. -Chad
+
+  def x_distance(x) #calculates how many tiles piece has moved on x-axis
     (x_position - x).abs
   end
 
-  def y_distance(y) #how many tiles piece has moved on y-axis
+  def y_distance(y) #calculates how many tiles piece has moved on y-axis
     (y_position - y).abs
   end
 
-  def moved?(x,y) #validate whether piece is at its origin position
+  def moved?(x,y) #validates whether piece is at its origin position
     x != x_position || y != y_position
   end
 
-  def horizontal_or_vertical_move?(x,y) #validate a move any amount of spaces along one axis
+  def horizontal_or_vertical_move?(x,y) #validates a move any amount of spaces along one axis
     (x_distance(x) == 0 || y_distance(y) == 0)
   end
 
-  def diagonal_move?(x,y) #validate a move any amount of spaces along a diagonal
+  def diagonal_move?(x,y) #validates a move any amount of spaces along a diagonal
     x_distance(x) == y_distance(y)
   end
 
-  def radial_move?(x,y) #validate a move 1 space in any direction
+  def radial_move?(x,y) #validates a move 1 space in any direction
     x_distance(x) < 2 && y_distance(y) < 2
   end
 
-  def forward_move?(x,y) #validate a move 1 space forward
+  def forward_move?(x,y) #validates a move 1 space forward
     x == x_position + 1 if color == 'black'
     x == x_position - 1 if color == 'white'
   end
 
-  def first_forward_move?(x,y) #validate a move 1-2 spaces forward
+  def first_forward_move?(x,y) #validates a move 1-2 spaces forward
     y == y_position && (x == x_position + 1 || x == x_position + 2) if color == 'black'
     y == y_position && (x == x_position - 1 || x == x_position - 2) if color == 'white'
+  end
+
+  def l_shaped_move?(x,y) #validates a move in an L-shape
+    (x_distance(x) == 2 && y_distance(y) == 1) || x_distance(x) == 1 && y_distance(y) == 2)
   end
 
 end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -35,13 +35,13 @@ class Piece < ActiveRecord::Base
   # Returns true if the coordinates provided
   # have the same x-axis value.
   def horizontal_move?(x,y)
-    x_distance(x) == 0
+    y_distance(y) == 0
   end
 
   # Returns true if the coordinates provided
   # have the same y-axis value.
   def vertical_move?(x,y)
-    y_distance(y) == 0
+    x_distance(x) == 0
   end
 
   # Returns true if the coordinates provided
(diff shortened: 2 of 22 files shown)


app/models/bishop.rb
```diff
@@ -1,2 +1,7 @@
 class Bishop < Piece
+
+  def valid_move?(x, y) #NO CAPTURE/COLLISION LOGIC YET
```
lasley: I like what you were going for here with the inline comments. Here's my suggestion if you do actually want to make your code look badass and readable:

Inline comments are typically unnecessary. The code should speak for itself in the sense that your variable/method names should all be verbose and the logic should not nested deep in data structures. It should be assumed that any developer reading a method and understand the fairly small amount of lines in it.

What is important is called the Doc Block. A Doc Block describes what a method does, what arguments (types and definitions) it accepts, and what is returned. This helps developers understand your method in relation to all the code without having to read the method.

The reason one is important (IMO) and not the other is that if you are in the context of a function, you are already in the context of the function. You've already likely read the code leading up to where you are, and you completely understand the variables that are in play.

When you're in a different method though, you don't want to have to read another method's code. You want to be able to get what you need (typically IDEs parse this automatically and provide in your context hints). 

http://guides.rubyonrails.org/api_documentation_guidelines.html

app/models/piece.rb
```diff
@@ -5,7 +5,52 @@ class Piece < ActiveRecord::Base
...
\ No newline at end of file
+  def diagonal_move?(x,y) #validates a move any amount of spaces along a diagonal
+    x_distance(x) == y_distance(y)
+  end
+
+  def radial_move?(x,y) #validates a move 1 space in any direction
+    x_distance(x) < 2 && y_distance(y) < 2
+  end
+
+  def forward_move?(x,y) #validates a move 1 space forward
```
lasley: This should go in Pawn unless something else is going to use it

app/models/piece.rb
```diff
@@ -5,7 +5,52 @@ class Piece < ActiveRecord::Base
...
+    if color == 'black'
+      y == y_position && x == x_position + 1
+    else
+      y == y_position && x == x_position - 1
+    end
+  end
+
+  def first_forward_move?(x,y) #validates a move 1-2 spaces forward
```
lasley: This should go in Pawn unless something else is going to use it

app/models/piece.rb
```diff
@@ -5,7 +5,52 @@ class Piece < ActiveRecord::Base
...
+    if color == 'black'
+      y == y_position && (x == x_position + 1 || x == x_position + 2)
+    else
+      y == y_position && (x == x_position - 1 || x == x_position - 2)
+    end
+  end
+
+  def l_shaped_move?(x,y) #validates a move in an L-shape
```
lasley: This should go in Knight unless something else is going to use it

app/models/piece.rb
```diff
@@ -5,7 +5,52 @@ class Piece < ActiveRecord::Base
...
-end
+  private
+
+  #ALL MOVE CHECKS ASSUME WHITE PLAYER IS AT BOTTOM OF 2-D ARRAY/BOARD, AND BLACK PLAYER IS ON TOP.  NO IMPACT ON VIEW. -Chad
+
+  def x_distance(x) #calculates how many tiles piece has moved on x-axis
+    (x_position - x).abs
+  end
+
+  def y_distance(y) #calculates how many tiles piece has moved on y-axis
```
lasley: What is the difference between this and `x_distance`?

app/models/piece.rb
```diff
@@ -5,7 +5,52 @@ class Piece < ActiveRecord::Base
...
+  end
+
+  def horizontal_or_vertical_move?(x,y) #validates a move any amount of spaces along one axis
+    (x_distance(x) == 0 || y_distance(y) == 0)
+  end
 
-# Sample piece creation - Piece.create(type:"Queen", color:"white") type must be capitalized.
-                      # - Rook.create(color:"white")
\ No newline at end of file
+  def diagonal_move?(x,y) #validates a move any amount of spaces along a diagonal
```
lasley: Could you possibly return either your x or y distance (known to be equal here), then just do the final eval in king? This would allow you to get rid of `radial_move`

app/models/queen.rb
```diff
@@ -1,2 +1,7 @@
 class Queen < Piece
+
+  def valid_move?(x, y)
+    moved?(x,y) && (radial_move?(x,y) || horizontal_or_vertical_move?(x,y) || diagonal_move?(x,y))
```
lasley: Is `radial_move` necessary here? Wouldn't this also be covered in a diagonal? If I'm reading correctly, we're evaling the x and y distance twice by calling both methods

app/models/bishop.rb
chadbaum: Got it.  Was generally aware that in-line commenting like that was a no-go in production but I wasn't sure the best way to inform the other team members of what was going on -- will take a look at Doc Block and explore converting to that.

app/models/piece.rb
chadbaum: So, I struggled with this as far as code organization.  The way I looked at it, I had some methods that were shared between the pieces related to their moves, and some that weren't.  But instead of mix and matching, I put ALL the move logic in Piece and then built the final validation function in the subclasses.  I was aware, and annoyed, the downside of this is that certain methods you expected to be in, for example, Pawn.rb, were now in Piece.rb, but I felt like I was choosing between two bad choices because you'd still have moments where you are not sure whether logic is in the subclass or parent class.

So given the ups and downs, I'm assuming the standard is to do what you suggested and only put stuff in Piece.rb that is shared by at least two subclasses, and deal with the tradeoffs?

app/models/piece.rb
chadbaum: I went back and forth on this one and I found that x_distance(x) and y_distance(y) were a bit easier to remember and write.  I just refactored it to distance(x, "x") and distance(y,"y") which is one function but takes in an additional parameter to look up the right home coordinate but I find it a bit more clunky to read.

chadbaum:
New changes incorporating @lasley comments.

aleksgorbenko:
I wasn't sure where to put one suggestion I have, but I think it is the most relevant place.

We need to add a validation to check that X and Y coords are actually within the boundaries of our 2D array. So what I propose is to add the following method to the Piece.rb model:

`def valid_coords?(x, y)
    (x <= 7 && x >= 0) && (y <= 7 && y >= 0)
end`

As it will return boolean - we just stick this method into the `valid_move?` method for each Piece.

So for example, Knight.rb would look like that:

`def valid_move?(x,y) #DOES NOT INCLUDE CAPTURE LOGIC YET
    moved?(x,y) && l_shaped_move?(x,y) && _valid_coords?(x, y)_
 end`

I think this logic could be implemented with the current PR.

chadbaum:
This will be added to the PR as per our discussion.  Thanks!

chadbaum:
Actually the more that I think about it, I am not sure this is necessary.  I was going to implement it as an ArgumentError if x or y is outside the parameters (above 7, below 0).  But then I realized there is no danger of this happening by the user.  The user is going to either drag and drop or have valid click boxes tied to our array.  There won't be any way for the user to tell a piece to go into an x,y coordinate other than what we present to them through the front-end.  Does this make sense?  If so, I think this code actually is not needed.

chadbaum:
This is ready to be code reviewed.  Removed WIP tag.

chadbaum:
I realized after reviewing Aleks's code that I swapped the X and Y axis.  Will fix and re-comment later.

lasley:
LGTM 👍

aleksgorbenko:
I agree that the user technically shouldn't be able to put the figure outside of the chessboard, I would still implement this logic just to make sure we are not having any possibility of that.

Having said that, if the majority thinks it is not necessary - we can skip it.

Apart from that - LGTM (just googled what it means hehe)

chadbaum:
I'm going to merge this for now since it has two thumbs up.  As for the out-of-bounds validation, I still believe it should not be added because there should be no scenario where the code is necessary, and I don't think unnecessary code should be implemented.  But we can make a new branch and PR if team wants to add it.  Let's discuss at next meeting.

aleksgorbenko:
Yep - agreed, we can bring it up on a session.

PR #25, opened 2016-07-13
Populate --Initial Board Population

GeorgeFourikis:
Hello there,so i added a model named Game.rb which is in relation to the Piece model(and the rest subclasses there).i ve built a method to create the Types  that we want,the specific piece types are found in the Piece model made by Zach and Chad i think.Feel free to check this out.

PS:I spent 1 full-day not having right results cause i was not including the specific color (string) value for each type(piece)!!


aleksgorbenko:
Not entirely sure, probably worth checking with @lasley , but `id` in the Game model should be with a colon? like that ---> `:id` ?

lasley:
@aleksgorbenko - I'm not sure where you're referring to. Could you drop an inline comment on the specific line instead, so I can get some context?

chadbaum:
I think you should add several tests before this code can be merged.

lasley:
Agreed. No logic should be written without an associated test.

Logic is defined as:
* Equations
* Variable assignments
* Evaluations
* Loops
* Method calls

Which basically means every line of code ;)

app/models/game.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def populate_board!
+
+
+    # Black Pieces of the board
+    (0..7).each do |i|
+      Pawn.create(game_id: id, x_position: i, y_position: 1, color: "black")
+    end
+
+    Rook.create(game_id: id, x_position: 0, y_position: 0, color: "black")
```
aleksgorbenko: @lasley I meant over here - for each Peace.create in the params, shouldn't it be `game_id: :id` ?

chadbaum: It does not need to be a symbol but I don't believe id will work anyway as 'id' will probably be an undeclared variable.  The problem here is that the pieces created here need to belong to the game.  Look at the rails associations that were set up for Game and Piece and use different syntax to create each piece so that it automatically belongs to the game.  The correct syntax should be `game.pieces.create(type: 'Rook', color: 'black, x_position: 0, y_position: 0)'.  This will also be revealed when a suite of unit tests are added.

db/migrate/20160712160707_create_games.rb
```diff
@@ -0,0 +1,16 @@
...
+      t.integer :white_player_id
+      t.integer :black_player_id
+      t.integer :winning_player
+
+
+
+      t.timestamps null: false
+    end
+  end
+end
```
chadbaum: Can you explain some of the decisions made here vs. the model spreadsheet:

https://docs.google.com/spreadsheets/d/1hijYjebn6_fSKAK5FeJLfVqfIGHHE51-iRgSyJNuEDQ/edit#gid=0

spec/models/game_spec.rb
```diff
@@ -0,0 +1,4 @@
+require 'rails_helper'
+
+RSpec.describe Game, type: :model do
+end
```
chadbaum: Please add several unit tests here.

spec/factories/games.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :game do
+    
+  end
+end
```
chadbaum: You will probably want to make a few factories for your unit tests here.

app/models/game.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def populate_board!
+
+
+    # Black Pieces of the board
+    (0..7).each do |i|
+      Piece.create(type: 'Pawn', game_id: id, x_position: i, y_position: 1, color: "black")
+    end
+
+    Piece.create(type: 'Rook',game_id: id, x_position: 0, y_position: 0, color: "black")
```
chadbaum: Hey, please see the comment on best way to do this using rails associations.

db/schema.rb
```diff
@@ -54,8 +54,8 @@
     t.datetime "updated_at",                          null: false
   end
 
-    t.index ["email"], name: "index_users_on_email", unique: true, using: :btree
-    t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true, using: :btree
+  add_index "users", ["email"], name: "index_users_on_email", unique: true, using: :btree
+  add_index "users", ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true, using: :btree
   
 
 end
```
chadbaum: To be clear, you should **never**  be directly editing the schema.rb file.  It is a dump for all the changes you should be making through valid migrations.  If you have been doing this, please revert all changes and do it through migrations or we will have no valid migration tree to undo anything you've done or change things.

app/models/game.rb
```diff
@@ -1,46 +1,37 @@
+# model Game,piece creation inside
+# rubocop:disable Metrics/AbcSize
 class Game < ActiveRecord::Base
   has_many :pieces
   has_many :players
 
   after_create :populate_board!
 
+  # rubocop:disable MethodLength
```
chadbaum: Don't use this.  Just think about how you can refactor your code to make Rubocop pass.  Hint:  break this into smaller methods.

app/models/game.rb
```diff
@@ -1,46 +1,37 @@
+# model Game,piece creation inside
+# rubocop:disable Metrics/AbcSize
```
chadbaum: Don't use this.

aleksgorbenko:
Ok we have shortened the method, Rubocop is not angry anymore. I think we are good to go.
👍

spec/models/game_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe Game, type: :model do
+  describe 'populate the board' do
+    it 'should give us 32 pieces upon board population' do
+      g = FactoryGirl.create(:game)
```
lasley: There's a lot of repetition with the game creations that could be eliminated by instead using a `let` statement to create/assign the game outside of your describe blocks.

This will also have an added benefit of speeding up your tests. As it is right now the game is being created 5 times, which means we have methods running upwards of 80 times as can be seen by the 80s to the right of the pawn creations in Coveralls - https://coveralls.io/builds/7087197/source?filename=app%2Fmodels%2Fgame.rb#L35

lasley:
Looks good. One comment on the tests, but I won't block for it. I would recommend fixing though.

GeorgeFourikis:
So i changed the creation of the game inside the tests!

chadbaum:
:+1:

PR #31, opened 2016-07-15
Move Logic for Pieces model

chadbaum:
Implemented move logic for Pieces.model which is inherited by all subpieces and provided a block comment to explain functionality.  Included rspec tests for valid and invalid moves for all subpieces.

### DIFF
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -5,16 +5,16 @@ class Piece < ActiveRecord::Base
   validates :color, inclusion: { in: %w(black white) }
   validates :type, inclusion: { in: %w(Pawn Rook Bishop Knight King Queen) }
 
-  # Returns "valid" and updates the piece's x and
+  # Returns true and updates the piece's x and
   # y position to provided coordinates if move is valid,
-  # otherwise do nothing and return nil.
-  def move(x,y)
+  # otherwise do nothing and return false.
+  def move!(x,y)
     if valid_move?(x,y)
       self.x_position = x
       self.y_position = y
-      return "valid"
+      return true
     else
-      return nil
+      return false
     end
   end
 
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -5,6 +5,19 @@ class Piece < ActiveRecord::Base
   validates :color, inclusion: { in: %w(black white) }
   validates :type, inclusion: { in: %w(Pawn Rook Bishop Knight King Queen) }
 
+  # Returns "valid" and updates the piece's x and
+  # y position to provided coordinates if move is valid,
+  # otherwise do nothing and return nil.
+  def move(x,y)
+    if valid_move?(x,y)
+      self.x_position = x
+      self.y_position = y
+      return "valid"
+    else
+      return nil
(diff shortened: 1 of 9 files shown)


app/models/piece.rb
```diff
@@ -5,6 +5,19 @@ class Piece < ActiveRecord::Base
...
   validates :type, inclusion: { in: %w(Pawn Rook Bishop Knight King Queen) }
 
+  # Returns "valid" and updates the piece's x and
+  # y position to provided coordinates if move is valid,
+  # otherwise do nothing and return nil.
+  def move(x,y)
+    if valid_move?(x,y)
+      self.x_position = x
+      self.y_position = y
+      return "valid"
```
lasley: Returning a string is weird & you will now have to remember what string you actually chose for valid. Typically when you do something like this, you would want to use a CLASS_LEVEL_CONSTANT to define the string, then use that everywhere.

Why not just return true instead though?

app/models/piece.rb
```diff
@@ -5,6 +5,19 @@ class Piece < ActiveRecord::Base
...
+    else
+      return nil
```
lasley: I'd return False here. `nil` is an indicator of nothing. By contrast, this is in fact something - a negative move case.

app/models/piece.rb
chadbaum: I agree, but I didn't want to make it a boolean method because the return value is not the main feature of the method.  I figured the really important thing was that it would return nil if it failed.  What's the classiest way to do this?

app/models/piece.rb
chadbaum: See above note as well.

Can I make it return true and false but not put a huh operator as the method name, because I think that is a bit misleading as to the nature of the method.  Or is making it a huh operator the way to go even though its main feature would be its 'side effect'?

app/models/piece.rb
lasley: Well it doesn't necessarily need to be a bang method, but I would say that a true/false return indicating whether the move actually happened is completely valid to expect from this method.

This is one of the things that bugs me with rails. Technically this should be a bang method. But it should also be a huh. But really, why the hell did they decide we had to decide? What the crap is wrong with a regular method... /rant

chadbaum: I wasn't sure on the bang, so clearly I need to get a finer grained appreciation for that distinction.  I'lll make the executive decision to go back to having it return true/false, but keep it as a bang operator.  I think having the huh operator makes the method unclear as to its primary purpose and is why I went down this road of crapola 'valid'/nil in the first place as a misguided attempt to find a middle ground that was functional and stylistic.  I failed. :)

lasley: Exactly. With a bang, you could technically infer that if the method doesn't raise an exception and break the program things are valid.

I don't really like that though, because sometimes having the boolean is just plain useful. Typically I'm for better to ask for forgiveness than permission, but sometimes both is good.

lasley:
LGTM aside from the comment 👍

Se7enB2st:
👍

PR #33, opened 2016-07-17
Add capture logic

Se7enB2st:
Not including which game pieces belongs to (still waiting for game model)
Not including collision logic

### DIFF
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -31,12 +31,13 @@ class Piece < ActiveRecord::Base
       end
     else
       return false
-    end  
+    end
   end
 
+  # if piece is capturable change piece's attributes
   def capture!(x,y)
     if capturable?(x,y)
-      target_piece(x,y).update_attributes(x_position: nil, y_position: nil, captured: true)
+      target_piece(x,y).update_attributes!(x_position: nil, y_position: nil, captured: true)
     end
   end
 
@@ -47,9 +48,9 @@ class Piece < ActiveRecord::Base
     false
   end
 
-  # Find the target piece base on game id, and x,y coordinates
+  # Find the target piece base on it x and y coordinates
   def target_piece(x,y)
-    Piece.find_by(game_id: game_id, x_position: x, y_position: y)
+    game.pieces.find_by(x_position: x, y_position: y)
   end
 
   # All validation assumes white player is on the
@@ -57,6 +58,11 @@ class Piece < ActiveRecord::Base
   # 0-1 rows of the array.
 
   private
+
+  # check if position is taken by any piece
+  def position_taken?(x,y)
+    game.pieces.find_by(x_position: x, y_position: y).present?
+  end
 
   # Compares a piece's x_position with the
   # coordinate provided and returns the
@@ -96,10 +102,5 @@ class Piece < ActiveRecord::Base
   # origin point along both axis.
   def diagonal_move?(x, y)
     x_distance(x) == y_distance(y)
-  end
-
-  # Return true if target coordinates is taken base on game id and x,y coordinates
-  def position_taken?(x,y)
-    Piece.find_by(game_id: game_id, x_position: x, y_position: y).present?
   end
 end
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -1,5 +1,4 @@
 # Piece superset behavior.
-require 'pry'
 class Piece < ActiveRecord::Base
   belongs_to :game
   belongs_to :player
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -38,8 +38,7 @@ class Piece < ActiveRecord::Base
   # Return true if target is a friendly piece
   # otherwise return false
   def friendly_piece?(x, y)
-    return true if color == target_piece(x, y).color
-    false
+    color == target_piece(x, y).color
   end
 
   # Find the target piece base on it x and y coordinates
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -17,28 +17,22 @@ class Piece < ActiveRecord::Base
     true
   end
 
-  # First check if coordinates is taken
-  # return false if coordinates belongs to friendly piece
+  # Return false if coordinates is taken but belongs to friendly piece
   # otherwise return true
-  # return false if position is not taken
   def capturable?(x, y)
-    if position_taken?(x, y)
-      return false if friendly_piece?(x, y)
-      true
-    else
-      false
-    end
+    return false unless position_taken?(x, y) && !friendly_piece?(x, y)
+    true
   end
 
   # if piece is capturable change piece's attributes
   def capture!(x, y)
-    if capturable?(x, y)
-      target_piece(x, y).update_attributes!(
-        x_position: nil,
-        y_position: nil,
-        captured: true
-      )
-    end
+    return false unless capturable?(x, y)
+    move!(x, y)
+    target_piece(x, y).update_attributes!(
+      x_position: nil,
+      y_position: nil,
+      captured: true
+    )
   end
 
   # Return true if target is a friendly piece
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -21,34 +21,35 @@ class Piece < ActiveRecord::Base
   # return false if coordinates belongs to friendly piece
   # otherwise return true
   # return false if position is not taken
-  def capturable?(x,y)
-    if position_taken?(x,y)
-      if friendly_piece?(x,y)
-        return false
-      else
-        return true
-      end
+  def capturable?(x, y)
+    if position_taken?(x, y)
+      return false if friendly_piece?(x, y)
+      true
     else
-      return false
+      false
     end
   end
 
   # if piece is capturable change piece's attributes
-  def capture!(x,y)
-    if capturable?(x,y)
-      target_piece(x,y).update_attributes!(x_position: nil, y_position: nil, captured: true)
+  def capture!(x, y)
+    if capturable?(x, y)
+      target_piece(x, y).update_attributes!(
+        x_position: nil,
+        y_position: nil,
+        captured: true
+      )
     end
   end
 
   # Return true if target is a friendly piece
   # otherwise return false
-  def friendly_piece?(x,y)
-    return true if color == target_piece(x,y).color
+  def friendly_piece?(x, y)
+    return true if color == target_piece(x, y).color
     false
   end
 
   # Find the target piece base on it x and y coordinates
-  def target_piece(x,y)
+  def target_piece(x, y)
     game.pieces.find_by(x_position: x, y_position: y)
   end
 
@@ -59,7 +60,7 @@ class Piece < ActiveRecord::Base
   private
 
   # check if position is taken by any piece
-  def position_taken?(x,y)
+  def position_taken?(x, y)
     game.pieces.find_by(x_position: x, y_position: y).present?
   end
 
--- a/app/models/piece.rb
+++ b/app/models/piece.rb
@@ -1,4 +1,5 @@
 # Piece superset behavior.
+require 'pry'
 class Piece < ActiveRecord::Base
   belongs_to :game
   belongs_to :player
@@ -6,16 +7,49 @@ class Piece < ActiveRecord::Base
   validates :color, inclusion: { in: %w(black white) }
   validates :type, inclusion: { in: %w(Pawn Rook Bishop Knight King Queen) }
 
-  # Returns true, updates the piece's x and y position
-  # to provided coordinates if move is valid, and sets
-  # the moved flag to true otherwise do nothing and
-  # return false.
+  # Returns true and updates the piece's x and
+  # y position to provided coordinates if move is valid,
+  # otherwise do nothing and return false.
   def move!(x, y)
     return false unless valid_move?(x, y)
     self.x_position = x
     self.y_position = y
     self.moved = true
     true
+  end
+
+  # First check if coordinates is taken
+  # return false if coordinates belongs to friendly piece
+  # otherwise return true
+  # return false if position is not taken
+  def capturable?(x,y)
+    if position_taken?(x,y)
+      if friendly_piece?(x,y)
+        return false
+      else
+        return true
+      end
+    else
+      return false
+    end  
+  end
+
+  def capture!(x,y)
+    if capturable?(x,y)
+      target_piece(x,y).update_attributes(x_position: nil, y_position: nil, captured: true)
+    end
+  end
+
+  # Return true if target is a friendly piece
+  # otherwise return false
+  def friendly_piece?(x,y)
+    return true if color == target_piece(x,y).color
+    false
+  end
+
+  # Find the target piece base on game id, and x,y coordinates
+  def target_piece(x,y)
+    Piece.find_by(game_id: game_id, x_position: x, y_position: y)
   end
 
   # All validation assumes white player is on the
@@ -62,5 +96,10 @@ class Piece < ActiveRecord::Base
   # origin point along both axis.
   def diagonal_move?(x, y)
     x_distance(x) == y_distance(y)
+  end
+
+  # Return true if target coordinates is taken base on game id and x,y coordinates
+  def position_taken?(x,y)
+    Piece.find_by(game_id: game_id, x_position: x, y_position: y).present?
   end
 end
--- a/spec/models/piece_spec.rb
+++ b/spec/models/piece_spec.rb
@@ -1,7 +1,6 @@
 require 'rails_helper'
 
 RSpec.describe Piece, type: :model do
-
   describe 'moved flag for piece' do
     it 'should return false if piece has never moved' do
       bishop = FactoryGirl.create(:bishop, :white)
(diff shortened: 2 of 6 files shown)


app/models/piece.rb
```diff
@@ -18,6 +18,27 @@ def move!(x,y)
     end
   end
 
+  #Inital Capture Logic:
+  #Need game model to test id of game
+
+  #Set initial captured status to false
+  def default_captured_status
+    self.captured = false if self.captured = nil
+  end
```
chadbaum: We will just set the default value in the database for the captured flag to false.  No need to set it manually for every piece.  This method is not needed.

app/models/piece.rb
```diff
@@ -18,6 +18,27 @@ def move!(x,y)
...
+
+  #Find the target piece in the database base on the x,y coordinates
+  def capturable?(x,y)
+    Piece.where(:x_position => x, :y_position => y, :captured => false).where.not(:color => color).present?
```
chadbaum: `:x_position => x` has updated rails syntax logic and can be more simply expressed as `x_position: x`.

app/models/piece.rb
```diff
@@ -18,6 +18,27 @@ def move!(x,y)
...
+  end
+
+  #When the target is captured, change it's attributes and remove it from the board
+  def capture!(x,y)
+    opponent_piece = Piece.where(:x_position => x, :y_position => y)
+    if capturable?(x,y)
+      opponent_piece[0].update_attributes(:x_position => nil, :y_position => nil, :captured => true)
+    end
+  end
+
```
chadbaum: I would expect capturable? to be called in the move method, not here.  Captures are going to initiate from a player's desire to move a piece to an x,y coordinate, thus triggering the move method.

app/models/piece.rb
```diff
@@ -18,6 +18,27 @@ def move!(x,y)
...
+  #Find the target piece in the database base on the x,y coordinates
+  def capturable?(x,y)
+    Piece.where(:x_position => x, :y_position => y, :captured => false).where.not(:color => color).present?
+  end
+
+  #When the target is captured, change it's attributes and remove it from the board
+  def capture!(x,y)
+    opponent_piece = Piece.where(:x_position => x, :y_position => y)
```
chadbaum: This will need to be associated with game object, but also take a look at `find_by` vs `where`.  This code as it stands will return an activerecord association object instead of the piece itself.  You would need append a `take` or `first` method to it to actually return an object, but you can avoid that by using the `find_by` which expects to return a specific record.

app/models/piece.rb
chadbaum: Not withstanding the above comment, there is actually a bug here.

`if self.captured = nil` should be `if self.captured == nil'.   It's easy to mix up the assignment and comparison operators.

lasley: Test coverage is also missing for this method, which I recommend adding as well.

app/models/piece.rb
```diff
@@ -12,31 +12,33 @@ def move!(x,y)
     if valid_move?(x,y)
       self.x_position = x
       self.y_position = y
+      capture!(x,y)
```
chadbaum: The capturable method should be here validating whether or not capture! should ever be called first.

app/models/piece.rb
```diff
@@ -12,31 +12,33 @@ def move!(x,y)
...
 
-  #When the target is captured, change it's attributes and remove it from the board
-  def capture!(x,y)
-    opponent_piece = Piece.where(:x_position => x, :y_position => y)
-    if capturable?(x,y)
-      opponent_piece[0].update_attributes(:x_position => nil, :y_position => nil, :captured => true)
-    end
+  # Find the target piece base on game id, and x,y coordinates
+  def target_piece(x, y)
+    Piece.where(game_id: game_id, x_position: x, y_position: y).take
```
chadbaum: Use find_by instead of where.  It will return the piece instead of an activerecord association so you don't need to use take.

app/models/piece.rb
```diff
@@ -12,31 +12,33 @@ def move!(x,y)
...
-    self.captured = false if self.captured = nil
+  # First check if coordinates is taken
+  # return false if coordinates belongs to friendly piece
+  # return true if target piece's attributes changes
+  def capture!(x,y)
+    if position_taken?(x,y)
+      return false if friendly_piece?(x,y)
+      return true if target_piece(x,y).update!(x_position: nil, y_position: nil, captured: true)
+    end
   end
```
chadbaum: This is returning true/false for a non boolean method (capture!).  What I expected to see in the capture method is the actual capture -- where you change the properties of the victim piece.

app/models/piece.rb
```diff
@@ -12,31 +12,33 @@ def move!(x,y)
...
 
-  #Find the target piece in the database base on the x,y coordinates
-  def capturable?(x,y)
-    Piece.where(:x_position => x, :y_position => y, :captured => false).where.not(:color => color).present?
+  # Return true if target is a friendly piece
+  # otherwise return false
+  def friendly_piece?(x, y)
+    return true if color == target_piece(x,y).color
+    false
   end
```
chadbaum: You can write the same thing by just saying `color == target_piece(x,y).color`.  See implict returns in Ruby.

app/models/piece.rb
chadbaum: I don't see anywhere in the code where you are performing the capture.  `return true if target_piece(x,y).update!(x_position: nil, y_position: nil, captured: true)` returns true if the capture has happened already, but does not do the capture.

chadbaum:
I added some comments,  I think you need to think about what you are trying to implement.  You should be validating whether a capture is going to occur, and then perform that capture.  A validation is going to check whether the array is occupied by a hostile piece.  The capture logic is going to update that piece's attributes.  The move logic is already going to take care of the aggressor's piece going to that position.  I would also think about the edge case of the pawn, whom attacks diagonally but moves forward.

spec/models/piece_spec.rb
```diff
@@ -1,4 +1,36 @@
...
+      other_white_pawn = FactoryGirl.create(:pawn, :white, x_position: 4, y_position: 1, captured: false)
+
+      expect(other_white_pawn.capture!(white_pawn.x_position,white_pawn.y_position)).to eq false
+    end
+
+    it "return true if specified coordinates occupied by opponent's piece" do
+      black_rook = FactoryGirl.create(:rook, :black, x_position: 4, y_position: 0, captured: false)
+      white_pawn = FactoryGirl.create(:pawn, :white, x_position: 4, y_position: 1, captured: false)
+
+      expect(white_pawn.capture!(black_rook.x_position,black_rook.y_position)).to eq true
```
chadbaum: This test is not adequately testing whether your capture logic is working.  It's just testing whether or not you explicitly returned true.  There is no actual capture logic implemented in your code.

chadbaum: The test would be whether the victim piece is set to captured, and its x and y position are set to nil.  This will fail because nowhere in your code are you doing that.

spec/models/piece_spec.rb
```diff
@@ -13,4 +15,40 @@
...
+
+    it "return true if specified coordinates occupied by opponent's piece" do
+      black_rook = FactoryGirl.create(:rook, :black, :x_position => 4, :y_position => 0, :captured => false)
+      white_pawn = FactoryGirl.create(:pawn, :white, :x_position => 4, :y_position => 1, :captured => false)
+
+
+      expect(white_pawn.capturable?(black_rook.x_position,black_rook.y_position)).to eq true
+    end
+  end
+
```
Se7enB2st: Can't update attributes in the database for some reason, it works in the console.

chadbaum: Gonna need more information than that to help you troubleshoot.

app/models/piece.rb
```diff
@@ -18,12 +17,52 @@ def move!(x, y)
...
+    if position_taken?(x,y)
+      if friendly_piece?(x,y)
+        return false
+      else
+        return true
+      end
+    else
+      return false
+    end
+  end
```
chadbaum: This can be refactored if you think about combining your conditionals.

chadbaum:
Re-added WIP tag as your build is failing.

Se7enB2st:
Fixed Rubocop errors, should be good to go.

spec/models/piece_spec.rb
```diff
@@ -4,13 +4,140 @@
...
+  # before create a new game, destory all pieces pre-populated from game model
+  def empty_game
+    game = FactoryGirl.create(:game)
+    game.pieces.destroy_all
+    game.reload
+    game
+  end
```
chadbaum: Check into using rspec's `let` syntax to create variables that are shared among several tests.

app/models/piece.rb
```diff
@@ -18,12 +17,53 @@ def move!(x, y)
...
+  # otherwise return true
+  # return false if position is not taken
+  def capturable?(x, y)
+    if position_taken?(x, y)
+      return false if friendly_piece?(x, y)
+      true
+    else
+      false
+    end
+  end
```
chadbaum: This whole method can be refactored into `position_taken(x, y) && ! friendly_piece?(x, y)`.  Much simpler.

app/models/piece.rb
```diff
@@ -18,12 +17,53 @@ def move!(x, y)
...
+  # if piece is capturable change piece's attributes
+  def capture!(x, y)
+    if capturable?(x, y)
+      target_piece(x, y).update_attributes!(
+        x_position: nil,
+        y_position: nil,
+        captured: true
+      )
+    end
+  end
```
chadbaum: Can eliminate a line here by converting the conditional into a guard clause.  Guard clause would look like `return false unless capturable?(x, y)`.

spec/models/piece_spec.rb
```diff
@@ -4,13 +4,140 @@
...
+
       bishop.move!(3, 6)
+
       expect(bishop.moved).to eq true
     end
   end
+
+  describe 'capturable?' do
+    it 'return false if specified coordinates not occupied by any piece' do
+      game = FactoryGirl.create(:game)
```
chadbaum: Can you conform the language used for it blocks to the rest of the rspecs so they read consistently.  For example, your it statements should begin with 'should'.

app/models/piece.rb
chadbaum: How is capture! being called?  It is not integrated with the rest of the logic.

app/models/piece.rb
Se7enB2st: Your right, I will make this change now.

spec/models/piece_spec.rb
Se7enB2st: While 'let'  let us write DRY code in the RSpec test, it tend to hide a lot of things and tests can quickly grow out of control.

Here is a blog post from Thoughtbot about let: https://robots.thoughtbot.com/my-issues-with-let

chadbaum: Using let or before seems to be the standard.   I don't think our app is complicated enough to worry about what is being mentioned in a blog post.  We are not experienced enough to be wading into what is obviously a controversial topic by the author of the blog themselves:  "For a while now, we’ve recommended against using let in RSpec and minitest. We even have a blog post explaining our thoughts. Despite that, I still have trouble explaining my problems with let and before to other developers. They’re wonderful for cleaning up duplication, and can make your tests smaller. That has to be a win, right?

+
Maybe not."

I think we should be focused on best practices and the easiest way to determine what a best practice is as a junior developer is to look at what the industry does.

Se7enB2st: I always prefer readability over a marginal increase of DRY in the test, since I believe tests should act like a documentation for how the application functions. This is especially true for this case. Since we already use a callback to populate the chess board (which hides a lot of details already), if we use `let` or `before` it will hide further more details and make the tests even harder to understand and make refactoring in the future much more difficult. 

This is just my option on this subject, if you really want to use `let` or `before` I can implement it also.

lasley: I normally agree with explicit and readability over DRY, and totally agree with your thought process here @Se7enB2st 

Tests kind of break this rule in my eyes. There is a lot of code that is simply required for setup, which creates the need to break some best practices in the interest of DRY.

Let's take this game creation, for example. We are now adding a feature which will require a new argument when creating a new instance. Let's say the feature is whether you want to join the game as black or white, instead of the implicit black/white join logic we currently have.

In the above example, we now have to refactor every game creation in our code. If we used let, we're looking at maybe 1-2 edits. Without the use of edit, we're looking at 1^n edits, where n is the amount of tests we have involving a game.

All this said, design decision and I abstain.

lasley: Secondary note, and callout of great thinking - Tests are 100% your go to in terms of docs. They are never out dated ;)

Se7enB2st: I have been trying to implement `let` and even `before` in the tests. The tests work if you run them individually, but fails if you run them together. After some research, the reason was because of `skip_callback` method works in rails 4 but causing errors in rails 5. I am out of idea here, anyone knows how to solve this issue? 

Other people are having similar type of problem: https://github.com/lynndylanhurley/devise_token_auth/issues/500

chadbaum: I have already started implementing let for all of the model rspecs in my obstruction branch for the reason lasley mentioned -- we still have a lot of things to do with our models, including adding a join table for making a player black or white, so I want to create tests that will scale for that.  A vast majority of the model unit tests we have now are from my code so we might as well just use the same objects for your code.

Se7enB2st: That is fine, so should I refactor my tests after your objects are implemented or you have other suggestion?

chadbaum: I would just leave it as is, and I can add them to my PR for obstruction since I have had to refactor those tests anyway.

Se7enB2st: I was able to figure out last night. In rails - 5 I need to set up callback first in the factory, before skipping it.

app/models/piece.rb
```diff
@@ -18,12 +17,47 @@ def move!(x, y)
...
+  # if piece is capturable change piece's attributes
+  def capture!(x, y)
+    return false unless capturable?(x, y)
+    move!(x, y)
+    target_piece(x, y).update_attributes!(
+      x_position: nil,
+      y_position: nil,
+      captured: true
+    )
+  end
```
chadbaum: Instead of calling move in capture, I would integrate capturable/capture into move.  Here's the flow:

- Player wants to move his rook to 3,5.
- We call rook.move!(3,5).
- Move checks if a rook can make that move, otherwise reject move
- Move checks if the path is not obstructed between origin and destination, otherwise reject move
- Move checks that if there is a piece in the tile, that it is hostile, otherwise reject move
- Update coordinates of mover, and capture victim if there was one

Se7enB2st:
Add move! logic into capture logic
Add new test for capture logic
Add new game factory
Add `before` method in Rspec tests for scalability
Refactor code to make it more consistent

app/models/piece.rb
```diff
@@ -18,12 +17,47 @@ def move!(x, y)
...
+
+  # Return true if target is a friendly piece
+  # otherwise return false
+  def friendly_piece?(x, y)
+    return true if color == target_piece(x, y).color
+    false
+  end
```
chadbaum: This can be replaced by `color == target_piece(x, y).color`.  That statement itself will resolve to either true or false because of the logical comparison operator.  You don't need to explicitly return true or false.

Se7enB2st: Ya, you right : )

app/models/piece.rb
Se7enB2st: I think we shouldn't combine these 3 methods. Capture and move is a different concept if there is no piece on target coordinates, it would be moved not capture. Also, combine three methods into one method would decrease readability of code. Methods should not be long at all, each method should be transparently obvious and tells a story.

chadbaum: I think you are misunderstanding what I'm saying.  Move should initiate the capture method.  Move is the action that is cascading all these other actions.  The logic for move method is literally 1 line of code right now.  Adding a capturable condition to it that triggers capture will make it 2 lines of code.  The whole point is to have everything build into move in a way that is clear and transparent.

chadbaum: Worst case scenario just leave as is -- I'm going to have to refactor a chunk of the code anyway when I slot the big beast -- obstruction logic -- into it.  So I'm fine with not resolving this now.  Move, valid move, obstruction, and capture are going to merged anyway in my PR so let's not get hung up on this right now.

Se7enB2st: Ya, I agree. But after this experience, I feel like I need more practice dealing with the backend. Is there any logic that is not a blocker that I can work on now?

chadbaum: Since the meeting is tomorrow, let's discuss it then when new tickets are created.  In the meantime, the are plenty of tickets that you can and should grab from the backlog to work on that, in my opinion, are more urgent.

chadbaum:
:+1: I will refactor the Rspecs and capture logic after merge on my PR.

ZachBerman:
Looks good - will merge so Chad can have some fun

PR #34, opened 2016-07-18
Rubocop

aleksgorbenko:
Some lines I cannot shorten as the code breaks in this case, or some methods contain too much code. When I started today we had 142 errors, we are down to 21 now.


chadbaum:
Seems like this stuff is mostly just changing boilerplate stuff we didn't write in the first place so I wouldn't worry too much.  Going forward hopefully it'll make a meaningful difference in the code we actually write (I try to write my code to what I think is conforming to the style guide anyway).

lasley:
👎  Rubocop is still not running in the CI. There should be an explicit step in Codeship that indicates Rubocop is running, otherwise it is not

Valid Rubocop (step 11): 
image

http://stackoverflow.com/questions/15008751/how-to-integrate-rubocop-with-rake

lasley:
Note that it is running on master as step 14

chadbaum:
I am working on pushing commits to this branch to resolve many of the remaining conflicts so we can merge the other PR's that are now broken.

aleksgorbenko:
Tried several setups, its seems to be working now, but we need to reduce the errors, for now i got 92 left. Once again mostly are comment lines that are too long and missing top-level class specifications.

Basically, if we don't comply with rubocop CodeShip won't let us pass.

chadbaum:
Refactored all of the rspec and model code to be compliant with Rubocop.  Now you'll need to take all the remaining files that Rubocop is complaining about, which are Rails files, and exclude those directories because we don't need to worry about styling Rails or gem code.

chadbaum:
:1:

aleksgorbenko:
Alright all is passed - we are good to go.

lasley:
👍

PR #35, opened 2016-07-19
Add pry to gemfile

Se7enB2st:
Adding pry gem to development group


chadbaum:
Pry-rails is the gem I was referring to that adds pry to the rails console.  Pry is just pry itself which you can install locally if you want.

See here:  https://github.com/rweng/pry-rails

Se7enB2st:
Just update the gem to pry-rails, everything should be good now

lasley:
LGTM 👍

chadbaum:
:+1:

PR #36, opened 2016-07-19
Moved Flag on Piece Model

chadbaum:
Added a boolean flag for whether or not a piece has ever moved.  This will allow our pieces to properly support castling.

- Flag defaults to false and is changed to true upon every successful piece#move!
- Deleted antiquated first_move detection for king and pawn
- Refactored pawn to use new boolean flag in determining its valid move
- Updated rspec tests for valid move to use moved flag for pawn tests
- Created new rspec tests to test moved flag itself


lasley:
👍

ZachBerman:
LGTM

PR #39, opened 2016-07-20
[WIP] Draw pieces on board

aleksgorbenko:
All pieces are now drawn on the chess board. I added code to the same JS script that was used to created the board itself. Two loops to add the pawns and several 'append' jQuery methods for all other pieces. Data attributes were used to select the correct HTML elements.

- white are on top, black on the bottom
- added some styles to the application.scss

Right now it resembles the populate method in our Game model.

_Side note: SCSS for the chessboard will be moved to the separate file with the next drag_drop_ajax PR._

### DIFF
--- a/app/assets/javascripts/chessboard.js
+++ b/app/assets/javascripts/chessboard.js
@@ -1,10 +1,4 @@
-<div class="container">
-  <table id="chess-board" class="col-centered" >
-  </table>
-</div>
-
-<script>
-  $(function() {
+$(function() {
     // draw the chess board
     for (var y = 0; y <= 7; y++) {
         $("#chess-board").append(
@@ -84,4 +78,3 @@
         '<i class="white-pcs glyphicon glyphicon-king">');
 
     });
-</script>
--- a/app/assets/javascripts/chessboard.js
+++ b/app/assets/javascripts/chessboard.js
@@ -56,12 +56,12 @@ $(function() {
     // right side
     $(blackKing).append(
         '<i id="5" class="black-pcs glyphicon glyphicon-king">');
-    $(blackRookRight).append(
-        '<i id="6" class="black-pcs glyphicon glyphicon-tower">');
+    $(blackBishopRight).append(
+        '<i id="6" class="black-pcs glyphicon glyphicon-bishop">');
     $(blackKnightRight).append(
         '<i id="7" class="black-pcs glyphicon glyphicon-knight">');
-    $(blackBishopRight).append(
-        '<i id="8" class="black-pcs glyphicon glyphicon-bishop">');
+    $(blackRookRight).append(
+        '<i id="8" class="black-pcs glyphicon glyphicon-tower">');
 
 
     // draw black pawns
@@ -88,12 +88,12 @@ $(function() {
     // right side
     $(whiteKing).append(
         '<i id="29" class="white-pcs glyphicon glyphicon-king">');
-    $(whiteRookRight).append(
-        '<i id="30" class="white-pcs glyphicon glyphicon-tower">');
+    $(whiteBishopRight).append(
+        '<i id="30" class="white-pcs glyphicon glyphicon-bishop">');
     $(whiteKnightRight).append(
         '<i id="31" class="white-pcs glyphicon glyphicon-knight">');
-    $(whiteBishopRight).append(
-        '<i id="32" class="white-pcs glyphicon glyphicon-bishop">');
+    $(whiteRookRight).append(
+        '<i id="32" class="white-pcs glyphicon glyphicon-tower">');
 
     });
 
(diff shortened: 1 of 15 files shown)


chadbaum:
I would recommend that we consider drawing the board from the perspective of the player such that the player is always on the bottom.  We are not ready to fully-implement that yet, but something to think about as we move forward with this code.

In the alternative, as a backup, I would recommend you reverse where the white and black pieces are as that might confuse you when you work on any back-end function.  The back-end logic has the black pieces on the top, and the white pieces on the bottom.  Note though, that this is only with regard to logic, what the user sees is entirely up to us.

See here:  

https://drive.google.com/file/d/0B677qZss8C2zaE1hWTFZWHh6M2s/view

aleksgorbenko:
I will reverse the colours for now and yes - I have actually also thought of how we can colour the pieces and make them face appropriate player during the game session. 

As an idea - we can add `player_colour` column to our game table with a string parameter and then use ERB "if statement" on the game view with Jquery assigning/removing pre-defined classes.

Also - I can't access the file you linked to, requested a permission.

chadbaum:
We'll need to check if the active_user is a player in the game, and then simply draw the chess board in a way that the player's pieces are on the bottom of the board, regardless of whether they are the black or white player.

app/views/games/_chessboard.html.erb
```diff
@@ -19,6 +18,70 @@
...
+        '<i class="white-pcs glyphicon glyphicon-bishop">');
+    $(black_knights).append(
+        '<i class="white-pcs glyphicon glyphicon-knight">');
+    $(black_queen).append(
+        '<i class="white-pcs glyphicon glyphicon-queen">');
+    $(black_king).append(
+        '<i class="white-pcs glyphicon glyphicon-king">');
+
+    });
 </script>
```
chadbaum: Perhaps it makes sense to refactor this into smaller functions and/or move this to a JS file instead of using <script> given the length of this script for more readability?  Might not need a HTML partial afterwards.  You could also refactor many of these append functions into one function that takes in an additional parameter given they all do the same thing except change which glyphicon loads.

aleksgorbenko:
You are totally right - it is better to move into a separate file and remove the partial. With the shortening the code, i tried to think of a loop with append function, but the thing is I am appending different-coloured pieces to 2 sets of different coords. So I am not too sure how I can refactor it further.

app/assets/javascripts/chessboard.js
```diff
@@ -0,0 +1,80 @@
...
+            "<td data-x='3' data-y='" + y + "'></td>" +
+            "<td data-x='4' data-y='" + y + "'></td>" +
+            "<td data-x='5' data-y='" + y + "'></td>" +
+            "<td data-x='6' data-y='" + y + "'></td>" +
+            "<td data-x='7' data-y='" + y + "'></td>" +
+          "</tr>");
+    }
+
+    // draw black pawns
+    for (var i = 0; i <= 7; i++) {
```
lasley: What happens when the board is refreshed? Wouldn't we want to put the pieces on the board according to where they are in the current game instance?

lasley: Oops forgot to drop a link. This should put you on the right track http://stackoverflow.com/questions/18162197/rails-pass-data-to-javascript

app/assets/javascripts/game.js
```diff
@@ -42,7 +52,7 @@ $(function() {
         var gameId = currentUrl.substring(currentUrl.lastIndexOf('/') + 1);
         var thisCell = this;
         var pieceIdInDB = Number( droppedItem.attr('id') ) +
-                                ( (gameId - 1) * 32 )
+                                ( (gameId - 1) * 32 );
         $.ajax( {
           type: 'PATCH',
           url: '../games/' + gameId,
```
chadbaum: This comment is purely based on your commit title, I haven't checked the code specifically:

For the first iteration of getting our app up and running I would leave this logic to the back-end because it is already there.  We would only need to modify things to add more feedback (like if you pick up a piece, the view changes to show all of the valid or invalid spots -- this is something we should add LATER and instead focus on MVP).  So if we are not providing a dynamic cue, the only thing that you need to do from the front-end is call `move!(x, y)` from the piece being moved, and to the x-y coordinates of the space it's attempting to be dropped.  Then do error handling when the method returns false to inform the player to make another drag-drop, or refresh the board when the method returns true.  The method will handle everything else:  moved, valid move, obstruction, capture, and eventually check and/or check-mate.

I.e. the back-end will already return false if you try to capture your own piece, so don't reinvent the wheel. :)

aleksgorbenko:
Closing this PR as #40 uses most of what this PR has and even more. #39 is now redundant.

PR #40, opened 2016-07-21
Drag, Drop, AJAX functionality + Draw pieced on the board 

aleksgorbenko:
Ok so to make it work I had to base this PR on my previous PR #39 - draw pieces on chessboard.
Here is what was modified:
- installed jQuery gem, edited application.js and application.css.scss as per documentation
- edited the folder structure of the css - application is now mostly empty with @import rules, game.css is the new file where all the chessboard styles.
- updated the game.js file - it is simplified now, comments explain what each part of the code is doing. 

If you pull this branch drag-drop won't work correctly as AJAX will result in 404. If you want to see animation in action you will have to comment out the file like this:

```javascript
        // sends the new coords update to DB
        // $.ajax( {
        //   type: 'PATCH',
        //   url: 'games/game/id', // needs the currect url,
        //   dataType: 'json',
        //   data: {
        //     'x_position': pos_x,
        //     'y_position': pos_y
        //   }
        // })
        // If move is valid - removes draggable
        // element from old cell and appends to
        // new coordinates - 'this'.
        // .success( function() {
            $(droppedItem).detach().css({top: 0, left: 0}).appendTo(this);
        // });
```
Don't be surprised if you see several pieces in the same cell - validations do not apply at the moment as AJAX requests are not sent out and new x & y coords are not verified.


aleksgorbenko:
Latest updates are taken from #39, now the two PR requests are combined (as #40 is dependent on #39) and #39 can be removed.

Things that have been completed by now:

- each piece in the view has the ID attribute 1-32 to map to the Database IDs.
- `chessboard.js` and populate method are now mirroring each other.
- `game_spec.rb` and `game_factory.rb` are updated just to reflect the modified populate method. No logic has been changed.
- AJAX is now working - it sends the correct data to the server and picks up latest data from it.
- ATM update method in `games_controller.rb` is selecting the right Game, the right Piece and executing the move (no errors thrown), but the DB is not getting updated for some reason.
- if you drag any Piece, only the cells that it can be dropped on are highlighted.
- all file methods are commented and Rubocop compliant. 

I will keep working on the controller, and see if I can debug it.

aleksgorbenko:
OK, `move!` method now seems to work properly and DB is getting updated after the AJAX request. Some other business logic doesn't yet work. But for now, this PR can be reviewed, I am removing WIP tag.

app/assets/javascripts/chessboard.js
```diff
@@ -0,0 +1,103 @@
...
+            "<td data-x='5' data-y='" + y + "'></td>" +
+            "<td data-x='6' data-y='" + y + "'></td>" +
+            "<td data-x='7' data-y='" + y + "'></td>" +
+          "</tr>");
+    }
+
+    // assign variables to initial positions based on
+    // their data attributes
+    // black piece variables
+    var blackRookLeft = 'td[data-x="0"][data-y="0"]';
```
lasley: What if the pieces have moved from their initial positions & the page is being loaded?

app/assets/javascripts/chessboard.js
```diff
@@ -0,0 +1,103 @@
...
+    var whiteBishopRight = 'td[data-x="5"][data-y="7"]';
+    var whiteKnightRight = 'td[data-x="6"][data-y="7"]';
+    var whiteRookRight = 'td[data-x="7"][data-y="7"]';
+
+    // append BS glyphicons to each <td> elements with IDs
+    // matching the IDs in database
+    // draw black figures
+    // left side
+    $(blackRookLeft).append(
+        '<i id="1" class="black-pcs glyphicon glyphicon-tower">');
```
lasley: Shouldn't these IDs correspond to the piece ID in db, instead of hard coded?

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
+$(function() {
+    // Picks up all the latest positions from
+    // the database
+    var currentPositions = $('.piece_pos').data('positions');
```
lasley: May want to use an ID instead of a class selector here in order to guarantee a unique result. There can be two of the same class, but only one Id. jQuery selects anything that matches, so multiple matches would really screw with the intent.

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+
+    $.each(currentPositions, function(key, value) {
+      var currentCell = 'td[data-x="' + value.x_position + '"]' +
+                        '[data-y="' + value.y_position + '"]';
+      //formula to map DB pieace IDs to the IDs in view
```
lasley: s/pieace/piece

app/assets/javascripts/chessboard.js
lasley: NVM I see; lower

app/assets/javascripts/chessboard.js
```diff
@@ -0,0 +1,103 @@
...
+    // white piece variables
+    var whiteRookLeft = 'td[data-x="0"][data-y="7"]';
+    var whiteKnightLeft = 'td[data-x="1"][data-y="7"]';
+    var whiteBishopLeft = 'td[data-x="2"][data-y="7"]';
+    var whiteQueen = 'td[data-x="3"][data-y="7"]';
+
+    var whiteKing = 'td[data-x="4"][data-y="7"]';
+    var whiteBishopRight = 'td[data-x="5"][data-y="7"]';
+    var whiteKnightRight = 'td[data-x="6"][data-y="7"]';
+    var whiteRookRight = 'td[data-x="7"][data-y="7"]';
```
lasley: IMO you need a more dynamic data structure here. Lots of repetition, when something like (pseudo): `{white: {class: 'white-pcs', pieces: [{class: 'glyphicon-tower', id: 1, name: 'Rook'}], black: ...}` would work

lasley: (and then refactor below to kill the repetition too)

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+      var pieceIdOnBoard = value.id % 32;
+      $('#' + pieceIdOnBoard).appendTo(currentCell);
+    });
+
+    // Makes all cells that have <i> elements
+    // draggble. If move is invalid, the piece will
+    // return to its cell
+    $('#chess-board td i').draggable({
```
lasley: IMO force direct child on td `'#chess-board td > i'`

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+      opacity: 0.6,
+      revert: 'invalid'
+    });
+
+    // Each cell will accept only <i> elements.
+    // Only when pointer hovers over the cell
+    // the drop will occur.
+    $('#chess-board td').droppable( {
+      // white cannot be dropped on white and vice versa
+      // this logic can be removed when move! method works
```
lasley: IMO this method should stay even when move! works. Save server power with easy validations, why not?

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+      // properly
+      accept: function(item) {
```
lasley: Prefix all jQuery vars with `$` - `$item`

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+        if ( (item.hasClass('black-pcs') &&
+            $(this).children().hasClass('black-pcs') ) ) {
```
lasley: Inefficient IMO. Selector would be better: `$(this).children('i').hasClass('black-pcs')`

lasley: (or whatever el the class would be in, I think it's an `i`)

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+          return false;
+        } else if ( (item.hasClass('white-pcs') &&
+            $(this).children().hasClass('white-pcs') ) ) {
```
lasley: Add selector

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+        // drop the piece
+        var posX = $(event.target).data('x');
+        var posY = $(event.target).data('y');
+
+        // get which item is being dragged
+        var droppedItem = ui.draggable;
+
+        // sends the new coords update to DB
+        var currentUrl = window.location.pathname;
+        var gameId = currentUrl.substring(currentUrl.lastIndexOf('/') + 1);
```
lasley: May want to strip possible suffix slashes (depending on how the routes are created, both /game/1 and /game/1/ are valid &
```
var currentUrl = '/game/1/'
// undefined
var gameId = currentUrl.substring(currentUrl.lastIndexOf('/') + 1);
// undefined
console.log(gameId)
// ""
```

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+        var thisCell = this;
+        var pieceIdInDB = Number( droppedItem.attr('id') ) +
+                                ( (gameId - 1) * 32 );
+        $.ajax( {
+          type: 'PATCH',
+          url: '../games/' + gameId,
```
lasley: Relative URIs are bad for many reasons. This article highlights the SEO reasons, but there are also security concerns that are a bit advanced - https://yoast.com/dev-blog/relative-urls-issues/

You don't need the domain, but you do need to start with a `/`. `/whatever/path/to/games/1`

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+        })
+        // If move is valid - removes draggable
+        // element from old cell and appends to
+        // new coordinates - 'thisCell'.
+        .success( function() {
+            $(droppedItem).detach().css({top: 0, left: 0}).appendTo(thisCell);
+        });
+      }
+    });
+});
```
lasley: https://robots.thoughtbot.com/no-newline-at-end-of-file

app/assets/javascripts/game.js
```diff
@@ -0,0 +1,77 @@
...
+        var posX = $(event.target).data('x');
+        var posY = $(event.target).data('y');
+
+        // get which item is being dragged
+        var droppedItem = ui.draggable;
+
+        // sends the new coords update to DB
+        var currentUrl = window.location.pathname;
+        var gameId = currentUrl.substring(currentUrl.lastIndexOf('/') + 1);
+        var thisCell = this;
```
lasley: JS practice is to call your `this` assignments `self`

app/models/game.rb
```diff
@@ -1,45 +1,54 @@
...
   has_many :players
 
-  after_create :populate_bishops!, :populate_rooks!, :populate_pawns!,\
-               :populate_knights!, :populate_queens_kings!
+  after_create :populate_left_black_half!,\
+               :populate_right_black_half!,\
+               :populate_black_pawns!,\
+               :populate_white_pawns!,\
+               :populate_left_white_half!,\
+               :populate_right_white_half!
```
lasley: Hmmm I'm not sure why I didn't comment on this initially, but why are we using all these methods? Wouldn't it be simpler to have one method for populate_board?

app/views/games/show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+
+  <hr class="col-xs-6 col-xs-offset-3">
+
+</div>
+<br>
+<br>
+<div class="container">
+  <table id="chess-board" class="col-centered" >
+  </table>
+</div>
```
lasley: https://robots.thoughtbot.com/no-newline-at-end-of-file

app/views/games/show.html.erb
```diff
@@ -0,0 +1,16 @@
+<br>
+<br>
+<%= content_tag :div, class: "piece_pos", data: { positions: @piece_positions } do %>
+<% end %>
+<div class="text-center players">
+  <h2>&#9819; Player1 vs &#9813; Player2</h2>
```
lasley: Should add the dynamic player names now. I don't see a ticket for it, and it's as simple as a few content_tags as above

app/assets/javascripts/chessboard.js
```diff
@@ -0,0 +1,103 @@
+$(function() {
```
lasley: Do we have test coverage to prove any of this works? 😉

app/assets/javascripts/game.js
aleksgorbenko: I originally intended it for readability. I can change it obviously to `$self` if that's the best practice.

app/models/game.rb
aleksgorbenko: We used to have one method, but then Rubocop was fussy about it, we then had populate_white and populate_black - still a no-go for Rubocop, so we had to split it into several methods.

app/assets/javascripts/chessboard.js
aleksgorbenko: We tested and every time the page loads it picks up current positions of pieces in DB and draws them on the board accordingly.

app/assets/javascripts/game.js
aleksgorbenko: Sorry @lasley I am confused with what you mean here :(

app/assets/javascripts/game.js
aleksgorbenko: just to confirm, are you referring to `accept` function?

app/assets/javascripts/game.js
lasley: Spelling issue, pieace should be piece

app/assets/javascripts/game.js
lasley: I was commenting on this comment ` // this logic can be removed when move! method works` and am referring to the plan of removing it

app/assets/javascripts/game.js
lasley: I'm pretty sure `this` is not a jQuery object, so it would simply be `self` but yeah best practice

app/models/game.rb
lasley: You have a link to the fails? Or remember the specific ones? I assume ABC?

aleksgorbenko:
So almost everything is fixed now as per @lasley comments.

Left to implement:

- User tags on the view (instead of Player1, Player2)
- tests are in the works, there are issues with the params of the `patch :update`params to test the updated coords of piece. As soon as the first test works more tests will be written to test different pieces.

app/models/game.rb
aleksgorbenko: Here is one: https://codeship.com/projects/160373/builds/16946103

Yes it was ABC.

spec/controllers/games_controller_spec.rb
```diff
@@ -21,9 +21,25 @@
...
     it 'should create a new game in the database' do
       process :create, method: :post, game: { turn: nil }
     end
   end
+
+  describe 'games#update' do
+    it 'should update the coords of the white pawn in the database' do
+      game = FactoryGirl.create(:game)
+      # test not working currently, something is wrong with its params. To be fixed.
+      patch :update, id: game.id, game: { piece_id: 24, x_position: 7, y_position: 5 }
```
aleksgorbenko: @lasley this is the line that we have the problem with. And this is the error that we get:

```Ruby
GamesController games#update should update the coords of the white pawn in the database
     Failure/Error: @piece = @game.pieces.find(params[:piece_id])

     ActiveRecord::RecordNotFound:
       Couldn't find Piece without an ID
```

chadbaum:
As per our discussion, :+1:

lasley:
Not going to refactor the piece adding into a data structure/loop?

aleksgorbenko:
At the moment the PR is big, we discussed it on a call, and we thought to merge it now to get it out of the way. We could refactor that at a later stage (I will make a ticket in the backlog).

lasley:
Works for me, I won't block

chadbaum:
Yeah we just had a long call to strategize how to knock out the remaining PR's and get the project moving forward and we felt that the best thing to do was to get this PR in and then work on PRs to refine it later on.  I already have one fixing CSS issues and refactoring the chessboard creation to a more efficient loop into an array into one dom manipulation (although I may want to kill that in favor of static code later on).

chadbaum:
Merging because Zach gave his :+1: on phone but forgot to add it here.

PR #41, opened 2016-07-25
Game model crud

ZachBerman:
-Add create, show, new actions in controller & add routes in routes.rb
-Authenticate user for new & create actions
-Add form for new game to new.html.erb
-Associated controller tests in game_controller_spec


app/controllers/games_controller.rb
```diff
@@ -1,13 +1,25 @@
 # Main controller for chess session for CRUD
 # logic for our app
 class GamesController < ApplicationController
+  before_action :authenticate_user!, only: [:create, :new]
+  def index
+  end
+
   def create
-    @game = Game.new
+    @game = current_user.game.create(game_params)
```
chadbaum: We didn't set up any associations with users and I don't see one in this PR so I would not expect this line to work.  There should also be a test for this because it would come back failing.

ZachBerman: This is a mistake and shouldn't be there - thought I removed it. Will do now

app/controllers/games_controller.rb
```diff
@@ -1,13 +1,25 @@
...
   end
 
-  def index
+  def show
+    @game = Game.new
   end
 
   def new
+    @game = Game.new
+  end
```
chadbaum: I don't foresee us using a new action but I think it's harmless to leave it for now.  We will probably have players create games straight from the index action (with either a 'Create Game' button or a modal form, depending on if and how we want to implement expanded actions the creator can take to customize the game.  I'd create a game object there instead.

chadbaum:
Other than the bug fix, and the comment about new action, I think this is good to go.  :+1:

Se7enB2st:
👍

PR #42, opened 2016-07-25
Obstruction Logic + Integrated Capture Logic

chadbaum:
Full obstruction logic for all pieces.

- Obstruction logic checks the tiles between origin and destination and rejects a move that has a piece in the way.  This logic does not apply to King or Knight regular moves.  It will apply to King castling but that is not implemented yet.
- Queen, rook, bishop, and the 2-forward-move of pawn all use the same logic in Piece.rb.  
- Pawn has an extra method for determining that the tile it is moving to is not occupied since this is an edge case where pawn cannot capture this piece if it used a forward-move.   This will be integrated with capture logic later.
- Move-set validation methods renamed to clear_?? to highlight that they now check for obstruction.  For example, `horizontal_move?(x, y)` is now `clear_horizontal_move?(x, y)`.  Note that these are all private helper methods and the only method that will be exposed to the front-end is `move!(x, y)` which will make this very easy to manage as we go forward.  Move will take care of everything.
- All unit tests for move validation needed to be rewritten.  I used rspec let variables where possible to reduce the code-base in rspec by about 40%.  
- Obstruction code was significantly refactored over time to be fully rubocop-compliant.

TO BE DONE in a later PR:

- Refactor database query technique for obstruction so that instead of making a query for each tile as its presented to the method, it takes the entire array of coordinates and queries ActiveRecord in one shot.
- Refactor Rspecs to re-integrate/make better use of FactoryGirl objects.

**I will have to do some significant refactoring to bring move, move validation, and obstruction logic together with Ethan's capture logic PR.  As a result, we will need to merge Ethan's code first, since its much smaller, and then I will handle the merge conflicts that result in this PR and then re-submit it.**


Se7enB2st:
👍

chadbaum:
I have fully integrated capture into `move!` and rewrote all the Rspecs to test the functionality.  I will perform further optimizations as discussed above in their own PR's at later dates as they are purely optional.

All methods have been flagged private except the `move!` method, as intended.

spec/models/knight_spec.rb
```diff
@@ -1,95 +1,103 @@
...
+      expect(knight.x_position).to eq 6
+      expect(knight.y_position).to eq 7
+      expect(knight.moved).to eq false
     end
+  end
 
-    it 'should return false if moved 1 space forward' do
-      knight = FactoryGirl.create(:knight, :black)
-      expect(knight.valid_move?(1, 1)).to eq false
+  describe 'invalid moveset' do
```
lasley: Please add a test for own piece in destination x,y

spec/models/king_spec.rb
```diff
@@ -1,95 +1,96 @@
...
+      expect(king.x_position).to eq 4
+      expect(king.y_position).to eq 7
+      expect(king.moved).to eq false
     end
+  end
 
-    it 'should return true if moved 1 space forward and 1 space right' do
-      king = FactoryGirl.create(:king, :black)
-      expect(king.valid_move?(4, 1)).to eq true
+  describe 'invalid moveset' do
```
lasley: Please add a test for own piece in destination x,y

app/models/pawn.rb
```diff
@@ -1,15 +1,17 @@
...
-  # Capture, collision, check, and checkmate logic are not
-  # implemented yet and thus ignored.
+  # move on its first move, or else a regular forward move,
+  # and ensure no capture takes place by moving forward.
+  # Check, and checkmate logic are not implemented yet
+  # and thus ignored.
   def valid_move?(x, y)
-    moved ? one_forward_move?(x, y) : first_forward_move?(x, y)
+    (moved ? one_forward_move?(x, y) : first_forward_move?(x, y)) &&
+      !forward_capture?(x, y)
```
lasley: What happens if there is an enemy directly in front of pawn on its first move, and pawn attempts a forward +2 move?

app/models/pawn.rb
```diff
@@ -30,17 +32,27 @@ def one_forward_move?(x, y)
...
   # are 1-2 tiles forward from the piece's origin.
   def first_forward_move?(x, y)
-    one_forward_move?(x, y) || two_forward_move?(x, y)
+    one_forward_move?(x, y) || clear_two_forward_move?(x, y)
+  end
+
+  # Returns true if there is any piece present
+  # in the tile provided.  Will be refactored when
+  # merged with capture logic.
+  def forward_capture?(x, y)
```
lasley: Seems like this could be squashed into valid_move; fairly useless method IMO

lasley: Also huh methods should be exclusively returning bool; this is a recordset

app/models/piece.rb
```diff
@@ -6,57 +6,43 @@ class Piece < ActiveRecord::Base
...
-    color == target_piece(x, y).color
+  # Returns true if position is occupied by a hostile piece.
+  def enemy?(victim)
+    color != victim.color
   end
 
-  # Find the target piece base on it x and y coordinates
-  def target_piece(x, y)
+  # Returns the piece occupying the coordinates.
+  def occupied?(x, y)
```
lasley: Huh methods should exclusively return booleans

app/models/piece.rb
```diff
@@ -79,21 +65,83 @@ def moved?(x, y)
...
+  # pieces in between.
+  def clear_diagonal_move?(x, y)
+    return false unless x_distance(x) == y_distance(y)
+    distance = x_distance(x)
+    path_clear?(x, y, distance)
+  end
+
+  # Class constants used to determine path direction
+  # for obstruction logic.
+  RIGHT = 1
```
lasley: Declare all constants at the top of your class

app/models/piece.rb
```diff
@@ -79,21 +65,83 @@ def moved?(x, y)
...
+  # Returns an array of x-y coordinate subarrays
+  # between origin and destination.
+  def generate_path_coordinates(x, y, distance)
+    coordinate_sets = []
+    direction = path_direction(x, y)
+    (distance - 1).times do |i|
+      i += 1
+      coordinate_sets << [
+        x_position + i * direction[:x],
+        y_position + i * direction[:y]
```
lasley: Even though here it doesn't truly matter due to just -1 or 1, had our numbers been slightly different you would have had some shitty results. Take a look at Ruby operator precedence (aka order of operations) - http://www.techotopia.com/index.php/Ruby_Operator_Precedence

lasley: Nope actually this is wrong; example:

```
2.1.2 :003 > x = 6
 => 6 
2.1.2 :004 > i = 6
 => 6 
2.1.2 :005 > d = -1
 => -1 
2.1.2 :006 > x + i * d
 => 0 
```

You want
```
2.1.2 :007 > (x + i) * d
 => -12 
```

lasley: When you fix this, we need test coverage for this bug as well.

app/models/piece.rb
```diff
@@ -79,21 +65,83 @@ def moved?(x, y)
...
+    end
+    coordinate_sets
+  end
+
+  # Returns true if no x-y coordinate pairs
+  # between origin and destination have a piece
+  # present.
+  def path_clear?(x, y, distance)
+    coordinates = generate_path_coordinates(x, y, distance)
+    coordinates.each do |coordinate|
```
lasley: Maybe look into `where` for array query - http://guides.rubyonrails.org/active_record_querying.html#subset-conditions

app/models/queen.rb
```diff
@@ -1,13 +1,13 @@
 # Queen behavior.
 class Queen < Piece
-  # Capture, collision, check, and checkmate logic are not
+  # Capture, check, and checkmate logic are not
   # implemented yet and thus ignored.
   def valid_move?(x, y)
     moved?(x, y) &&
       (
```
lasley: Would be better to use Enumerable.any? - http://apidock.com/ruby/Enumerable/any%3F

lasley:
Great job here, comments inline.

One general note - I noticed that you also cleaned up a lot of other code that wasn't relevant to this featureset. I am definitely all for cleaning up the code, but it would be best if that was done in an independent branch & PR in order to keep the diff cleaner for review.

app/models/pawn.rb
chadbaum: I forgot to wrap the truthy return of the recordset in true/false.  However, I realized there is a better way to do this directly with Activerecord queries through `exists?`.

chadbaum: And yeah, that was supposed to be eliminated as per the block comment.  I forgot. :)

chadbaum: Had trouble with ABC size errors with embedding it into the the valid move helper methods and I am loathe to make the parent valid_move method less readable so I kept it in a helper for now.

spec/models/knight_spec.rb
chadbaum: If you mean whether or not a piece is being moved from its original location, that is being tested here:  `  describe 'moved' do
    it 'should return false if not moved' do
      expect(knight.move!(6, 7)).to eq false
      expect(knight.x_position).to eq 6
      expect(knight.y_position).to eq 7
      expect(knight.moved).to eq false
    end
  end`

spec/models/king_spec.rb
chadbaum: See above, code is here:

`describe 'moved' do
    it 'should return false if not moved' do
      expect(king.move!(4, 7)).to eq false
      expect(king.x_position).to eq 4
      expect(king.y_position).to eq 7
      expect(king.moved).to eq false
    end
  end`

spec/models/knight_spec.rb
chadbaum: If you mean attempting to move onto a friendly piece, I put that test in the piece_spec as:

`    it 'should return false on a move against a friendly piece' do
      victim = game.pieces.find_by(x_position: 3, y_position: 6)
      expect(moved_queen.move!(3, 6)).to eq false
      expect(moved_queen.x_position).to eq 3
      expect(moved_queen.y_position).to eq 3
      expect(moved_queen.moved).to eq true
      victim.reload
      expect(victim.x_position).to eq 3
      expect(victim.y_position).to eq 6
      expect(victim.captured).to eq nil
    end`

spec/models/king_spec.rb
chadbaum: Test against friendly peace: 

`    it 'should return false on a move against a friendly piece' do
      victim = game.pieces.find_by(x_position: 3, y_position: 6)
      expect(moved_queen.move!(3, 6)).to eq false
      expect(moved_queen.x_position).to eq 3
      expect(moved_queen.y_position).to eq 3
      expect(moved_queen.moved).to eq true
      victim.reload
      expect(victim.x_position).to eq 3
      expect(victim.y_position).to eq 6
      expect(victim.captured).to eq nil
    end`

app/models/piece.rb
chadbaum: Changed to `exists?`

chadbaum: Nevermind, this needs to be changed from a ?.

app/models/piece.rb
chadbaum: I don't think this is correct.  The default order of operations here is what we want -- `direction`'s sole job is to flip `i` between being a positive and a negative.  If you place parenthesis on the first two elements it produces a completely different result.  What I'm using `direction` for is to determine whether to increment up or down on an axis.

The example above is actually bad data because if `x_position` is 6, 'i' says we're on our 6th iteration which means we're counting at least 6 tiles between origin and destination, and 'direction' is -1 which means we're going left, then we are going off the board.

Let's take the example of a queen at [6, 0] making a move to [6, 7].  The call to the function would be `generate_path_coordinates(6, 7, 7)`.  This represents the largest distance you could move across the board.

`[73] pry(main)> queen
=> #<Queen:0xbb8a2378
 id: 1059,
 color: "white",
 type: "Queen",
 x_position: 6,
 y_position: 0,
 game_id: 35,
 player_id: nil,
 captured: nil,
 checkmate: nil,
 created_at: Fri, 29 Jul 2016 13:41:38 UTC +00:00,
 updated_at: Fri, 29 Jul 2016 13:41:38 UTC +00:00,
 moved: true>
[74] pry(main)> queen.generate_path_coordinates(6,7,7)
=> [[6, 1], [6, 2], [6, 3], [6, 4], [6, 5], [6, 6]]
[75] pry(main)>
`

Here is the incorrect return values with the parens added:

`[7] pry(main)> queen
=> #<Queen:0xbaddb688
 id: 1092,
 color: "white",
 type: "Queen",
 x_position: 6,
 y_position: 0,
 game_id: 36,
 player_id: nil,
 captured: nil,
 checkmate: nil,
 created_at: Fri, 29 Jul 2016 14:03:51 UTC +00:00,
 updated_at: Fri, 29 Jul 2016 14:03:51 UTC +00:00,
 moved: true>
[8] pry(main)> queen.generate_path_coordinates(6,7,7)
=> [[0, 1], [0, 2], [0, 3], [0, 4], [0, 5], [0, 6]]
[9] pry(main)>
`

As for tests, there is test coverage for all of the helper methods but it's all wound up in the final result.  I read somewhere that testing specific helper methods was bad practice and that I should just test the final methods in a way that it would bubble up bugs with helper methods.  As an aside, all tests correctly break when changing parens. :)

Let me know if this makes sense or if I'm in outer space.

chadbaum:
Sorry, I tried to stick to things I was cleaning up that were related to my test coverage but in general I am keeping an eye on things I need to fix that belong in a later PR (or several).

app/models/queen.rb
chadbaum: I'm not sure how to go about this since I don't have a collection, just 3 method calls.  Is there a way to implement a collection of method calls that would make sense here to reduce complexity?

app/models/piece.rb
chadbaum: Should this go above associations and validations for ActiveRecord or below?

lasley: Above. I may be slightly off in regards to Rails styling, but here's a paste out of LasLabs styling guide:

In a Model attribute order should be

1. Private attributes (_name, _description, _inherit, ...)
2. Class constants
3. Fields declarations
4. Default method and _default_get
5. Compute, inverse, search, and group methods in the same order as field declaration
6. Constrains methods (@api.constrains) and onchange methods (@api.onchange)
7. CRUD methods (ORM overrides)
8. Action methods
9. And finally, other business methods.

You don't have 1, 4, and 6 for the most part.

spec/models/knight_spec.rb
lasley: Ok cool yup this actually worked itself out as I read the code, I just forgot to update the comments.

app/models/piece.rb
lasley: Nope I think I was just a bit tired; math makes much more sense now that I'm awake and have coffee. Disregard; although it still might be worthwhile to add parenthesis for clarity

app/models/queen.rb
lasley: Just wrap them in square brackets instead of parenthesis ;)

chadbaum: Cool, I actually tried that earlier in its own array just to test it, and Ruby couldn't parse it.  But it works now.

app/models/piece.rb
chadbaum: I can't figure this one out using the available methods for ActiveRecord.  Might need raw SQL.  Dunno.  But what I would need to be doing to make this work is use a method that tells the database to map the 0 index and 1 index of multiple arrays and map them to x_position and y_position.  I don't see how to make this work with `where` or better yet `exists?` using either hash conditions or array conditions.

chadbaum:
Minor changes are in based on @lasley's comments.

aleksgorbenko:
Ok, I went through it - all makes sense (not without the help of Chad :) ). I give 👍 , with a small suggestion of breaking up the Piece.rb into several files. Perhaps we can just create extra Module/s and require it/them in the Piece.rb model? 

1. `Piece.rb` could have just the move methods:

- `move!`
- `x_distance` and `y_distance`
- `moved?`
- `capture!`
- `enemy?`
- `occupant_piece`

`lib/distance_and_paths.rb` could have the rest:

- `clear_horizontal_move?`
- `clear_vertical_move?`
- `clear_diagonal_move?`
- `path_direction`
- `generate_path_coordinates`
- `path_clear?`

Or even make it more modular and create 2 modules. As it is at the moment having all 13 methods in one file is quite hard to read and understand. That's just IMHO.

If it is not a great idea/best practice, disregard and count this comment as 👍

app/models/piece.rb
lasley:
```
2.1.2 :008 > test, derp = [1, 2]
 => [1, 2] 
2.1.2 :009 > test
 => 1 
2.1.2 :010 > derp
 => 2 
2.1.2 :011 > 
```

lasley:
@aleksgorbenko I like your idea of splitting out the distance helper library, however the `path_clear?` method would require some deep abstraction in order to make it work & not be fully dependent on the game. 

Not impossible & definitely a good architecture decision if we were going to, say, build a Checkers game too. Typically this type of abstraction is tackled when the second project comes up though, because abstracting existing code is easier and more maintainable than creating anew. Trying to abstract everything from the beginning can become a rabbit hole IMO

lasley:
👍

ZachBerman:
Look good! Merging now

PR #48, opened 2016-07-31
[#62] [WIP] En Passant Logic For Pawn

Se7enB2st:
-Dependent on [Turn Base Logic PR](https://github.com/TeamStoic/StoicChess/pull/62)

This is building off Chad's obstruction PR. 

Add:
- `first_turn!` to game model
- `end_turn!` to game model
- `En Passant logics` to pawn model
- `move_number` to game database table
- `en_passant` to game database table
- `last_moved_piece` to piece database table

Not including:
- Pawn Capture Logic 
- En Passant Capture Logic (Dependent on Pawn Capture Logic)

En Passant Logic:

 1) Target pawn just moved 2 space forward
 2) Target pawn only capturable within the next turn
 3) Target pawn can only be captured by pawn of opposite color
 4) Opposite Pawn must be adjacent to the Target pawn ( `x+1` or `x-1`)

There is probably some refactoring can be done, feel free to leave feedbacks.


lasley:
Rebased onto master for branch isolation

chadbaum:
I suggest we merge pawn capture logic first so Ethan can then complete en passant logic as noted in desc.

chadbaum:
Now that pawn capture is in, pls confirm work is done and PR is ready for review by removing WIP tag and commenting.

Se7enB2st:
Integrated En Passant Logic into Pawn's valid_move? logic. En Passant Logic now works with piece move! logic.

lasley:
Looks like we have conflicts

Se7enB2st:
Fixed Merge Conflicts, also add a test to make sure En Passant Logic only works on enemy pawn.

chadbaum:
Can you explain your schema decisions?  Scanning your code, you have added:

Piece:  move_number, turn, last_moved_piece
Game:  en_passant

As far as I can tell:

- En_passant is not used for anything in your code, please delete or explain.
- As discussed at last 2 meetings, `move_number` and `turn` are both redundant, you only need one.  An odd turn means it is white player's turn, an even turn means it is black player's turn.  There is no need for another column here.  I would delete `move_number` and just make `turn` an incrementing integer.
- I would rename last_moved_piece as it is both redundant (don't need to say piece when talking about piece model) but also vague...I think what you meant is something like `turn_last_moved`.

This also takes care of the turn logic in the next ticket so I'm not sure there is any work to be done there.  When this is done we can surface what turn it is in the games#show view with a simple `<%= game.turn %>'.

Also, I noticed a bunch of surface-level changes to files unrelated to en passant, do you know what caused them and if we should be reviewing any of it?  If you look at your files changed, you have many initializers, application.rb, and environment files included.  Want to make sure you're not wasting your time Rubocopping any of these files -- they should have already been blacklisted.

Se7enB2st:
En Passant is not needed at this point(It was necessary when I first attempt at En Passant Logic), I will clean up the database to remove/change any redundant columns.

The surface-level changes were due to Rubocop errors. It kept asking me to add top-level comments at those files(Not sure why it was doing this if it was blacklisted). But there is no other change made to those files beside the top comments.

chadbaum:
@aleksgorbenko Do you know why Ethan running Rubocop would have caused violations for these files when it doesn't for the rest of us?

@Se7enB2st I wouldn't waste your time Rubocopping Rails files.  I've not had any problems with the whitelist/blacklist since Aleks implemented it so maybe he can investigate.

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
...
+  end
+
+  # Check for any adjacent piece
+  # Return false if none if found
+  # Return true if opponent's pawn is found
+  def valid_en_passant_pawn?(x, y, adjacent)
+    adjacent_piece = occupant_piece(x + (adjacent ? 1 : -1), y)
+    return false if adjacent_piece.nil?
+    return true if adjacent_piece.type ==
+                   'Pawn' && adjacent_piece.color != color
```
aleksgorbenko: i think for readability it should be:

```Ruby
return true if adjacent_piece.type == 'Pawn' &&
                      adjacent_piece.color != color
```

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
...
+  # En Passant Capture logic
+  def en_passant_capture?(x, y)
+    return true if en_passant_possible?(x, y) || en_passant_y_diff?(x, y)
+  end
+
+  # Return false if victim piece is nil or same color
+  # Return true  if victim piece's y axis difference is equal to 1
+  def en_passant_y_diff?(x, y)
+    victim = last_moved_piece
+    return false if victim.nil? || victim.color == color
```
aleksgorbenko: last moved piece will be always opponents piece, so `victim.color == color` validation seems to be unnecessary.

app/models/pawn.rb
Se7enB2st: Just tried, Rubocop gives me an error.  : (

aleksgorbenko:
@chadbaum re: Rubocop - no idea really. I just checked the rubocop.yml file and it is in place in this branch. Not sure what could have caused that.

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
 
   private
 
+  # En Passant Logic:
+  # 1)target pawn just moved 2 space forward
+  # 2)only capture within the next turn
+  # 3)can only capture by opponent's pawn
```
chadbaum: I'm not sure what step 3 means.  En Passant does not require that the ONLY piece available to capture the victim be the pawn, only that the pawn COULD have captured it in the first space on the previous turn.

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
...
+
+  # Return false if last move piece is nil (Start of game)
+  # Return false last moved piece is not pawn or its y axis difference is !2
+  # Check both both adjacent side of pawn
+  # If any have valid En Passant pawn piece, return true
+  def en_passant_possible?(x, y)
+    return false if last_moved_piece.nil?
+    return false unless last_moved_piece.y_distance(y - 2) == 2 &&
+                        type == 'Pawn'
+    [valid_en_passant_pawn?(x, y, true), valid_en_passant_pawn?(x, y, false)]
```
aleksgorbenko: this bit should be also made more readable

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
...
+    return true if  x == victim.x_position &&
+                    (y == victim.y_position - 1 || y == victim.y_position + 1)
+  end
+
+  # Return false if last move piece is nil (Start of game)
+  # Return false last moved piece is not pawn or its y axis difference is !2
+  # Check both both adjacent side of pawn
+  # If any have valid En Passant pawn piece, return true
+  def en_passant_possible?(x, y)
+    return false if last_moved_piece.nil?
```
chadbaum: Why do you need this guard clause if `last_moved_piece` returning nil in the next line will cause that to return false anyway?

aleksgorbenko:
I can't say I understand the methods entirely - it would be best to discuss on a call.

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
...
+    [valid_en_passant_pawn?(x, y, true), valid_en_passant_pawn?(x, y, false)]
+      .include?(true)
+  end
+
+  # Check for any adjacent piece
+  # Return false if none if found
+  # Return true if opponent's pawn is found
+  def valid_en_passant_pawn?(x, y, adjacent)
+    adjacent_piece = occupant_piece(x + (adjacent ? 1 : -1), y)
+    return false if adjacent_piece.nil?
```
chadbaum: Don't need this guard clause.

app/models/pawn.rb
Se7enB2st: This line is require for edge case, where at first turn there is no last moved piece.

app/models/pawn.rb
Se7enB2st: This method refactor version of following methods:
 
`def valid_en_passant_pawn?(x, y)
     return false unless valid_passant_pawn_left?(x, y)
     return false unless valid_passant_pawn_right?(x. y)
     false
  end
 
   def valid_passant_pawn_left?(x, y)
     left_piece = find_piece(x - 1, y)
     return false if left_piece.nil?
     return true if left_piece.type == 'Pawn' && side_piece.color != color
     false
   end
 
   def valid_passant_pawn_right?(x, y)
     right_piece = find_piece(x + 1, y)
     return false if right_piece.nil?
     return true if right_piece.type == 'Pawn' && side_piece.color != color
     false
   end`

app/models/pawn.rb
```diff
@@ -20,6 +20,49 @@ def promote!(new_type)
...
+  end
+
+  # Return false if last move piece is nil (Start of game)
+  # Return false last moved piece is not pawn or its y axis difference is !2
+  # Check both both adjacent side of pawn
+  # If any have valid En Passant pawn piece, return true
+  def en_passant_possible?(x, y)
+    return false if last_moved_piece.nil?
+    return false unless last_moved_piece.y_distance(y - 2) == 2 &&
+                        type == 'Pawn'
```
chadbaum: If I understand this correctly, you moved y_distance to a public method and are now invoking it on the last_moved_piece, which would be the pawn victim.  But if you pass in the coordinate of where pawn is going to "attack" as the coordinate behind where the victim is now, I don't see how this logic could work.  If victim is at 3,3 and aggressor pawn is at 4,3, calling this method on 3,2 which is where aggressor pawn would 'attack' would fail this check.   It would resolve as:

abs(victim.y_position - y)
abs(3 - 0)
abs(3)
3

app/models/pawn.rb
Se7enB2st: Tried to remove this line, but it causes the method to return Nil class error.

chadbaum: Can you explain what the adjacent parameter is doing?  I see you passing true and false between these 2 methods but they are not commented.

chadbaum: Just because a pawn is adjacent to another pawn does not make it a valid en passant victim so I think the name and return of this method are misleading.  What makes it valid is this PLUS the victim having moved 2 spaces forward last turn.

app/models/pawn.rb
chadbaum: I agree with Aleks, this should be refactored as he suggested for best formatting.  But it is minor.

app/models/pawn.rb
Se7enB2st: Since turn base logic is not implemented yet, this is needed at this point. Will refactor this method turn base logic is added in separate PR.

app/models/pawn.rb
chadbaum: I understand that, but why do you need this guard clause?  If the statement returns nil, it will be handled by next statement.

app/models/pawn.rb
Se7enB2st: The intent of this method is to pass the coordinates of where victim piece is at. 

I just refactor the method; now it works with `y_distance(y) == 2`. 
This method now will only pass if victim piece is pawn and moved two space forward in the last turn.

app/models/pawn.rb
Se7enB2st: It would cause an error in the next statement since you can't call any method on nil object.

app/models/pawn.rb
Se7enB2st: This method checks if there is any valid pawn adjacent to the victim pawn(include? true or false). If any of adjacent side have a valid pawn, it will return true.

This was the rafactored version of following methods:
```Ruby
def valid_en_passant_pawn?(x, y)
return false unless valid_passant_pawn_left?(x, y)
return false unless valid_passant_pawn_right?(x. y)
false
end

def valid_passant_pawn_left?(x, y)
left_piece = find_piece(x - 1, y)
return false if left_piece.nil?
return true if left_piece.type == 'Pawn' && side_piece.color != color
false
end

def valid_passant_pawn_right?(x, y)
right_piece = find_piece(x + 1, y)
return false if right_piece.nil?
return true if right_piece.type == 'Pawn' && side_piece.color != color
false
end
```

app/models/pawn.rb
chadbaum: Isn't turn-based logic completely implemented in this PR?  Turn in game model, start turn at 1, increment turn at end of every successful move.

Se7enB2st: It about 50% done in this PR, same color can move multiple turns in a row in this PR. I am working on the other half right now.

lasley:
This is good to review right, or are you guys still wokring on it?

Se7enB2st:
Yes, I will refactor more once turn base logic is merged.

app/models/pawn.rb
lasley: Guards should be as close to the error as possible so as to guarantee that the error is caught in the right spot & reduce the need for logic duplication.

app/models/pawn.rb
lasley: I'm confused by this, Is the turn logic being implemented in this PR then? We definitely require turn logic for a valid en passant

app/models/pawn.rb
lasley: I don't see the other method you are referring to that checks the 2 spaces forward. IMO this logic should all be one method, it is really hard to follow otherwise.

chadbaum:
This still does not seem ready to merge as there are many pending questions, so we will fix this for the next version release.

PR #49, opened 2016-08-01
Pawn Capture Logic

chadbaum:
This implements pawn capture in a way that plugs into the move method.  Pawn's valid move list has been expanded to accept a diagonal move forward either left or right but only if it is occupied by an enemy piece.  Then it's passed forward to the move method just like any other piece for move and capture processing.


spec/models/pawn_spec.rb
```diff
@@ -114,4 +114,47 @@
...
+      expect(moved_pawn.y_position).to eq 4
+      expect(moved_pawn.moved).to eq true
+      victim.reload
+      expect(victim.x_position).to eq 2
+      expect(victim.y_position).to eq 3
+      expect(victim.captured).to eq false
+    end
+
+    it 'should return true and update position on capturing enemy' do
+      victim = game.pieces.create(
```
aleksgorbenko: Could these variable creation be moved to the top using `let(:white_rook) do`?

chadbaum: Will refactor all the factorygirls (i feel sexist saying it that way but oh well) in sep PR

aleksgorbenko: Ok makes sense.

Se7enB2st:
👍

ZachBerman:
:+1:

PR #50, opened 2016-08-02
Bug Fix - Boolean Flag Defaults

chadbaum:
I noticed while implementing the capture/obstruction logic that the default for the captured flag was not set so that I had to check for nil instead of false.  I went in and changed the default values for the captured and the other boolean flag in our database already, checkmate, to default of false, as that is what we'd expect from a boolean, not nil.

Updated the rspec for capture logic to properly reflect this.


aleksgorbenko:
yep makes sense 👍

ZachBerman:
Looks good!

PR #52, opened 2016-08-04
Populate Game Fix

chadbaum:
I had several problems throughout the project conforming FactoryGirl to using the callbacks that we were using in populating the game board in the game model and it forced me to, at the time, discontinue using FactoryGirl for many things.  After consulting my mentor, and doing a ton of research on my own, while there are certain ways around the problem of callbacks in testing, it is actually considered bad practice to use them in situations here where they gain us nothing.

I have refactored the game model away from using callbacks and it now simply has one public method called ```populate!``` which will populate the board.  This makes RSpec testing much much easier both with and without FactoryGirl to set up full and empty games for testing.  I have updated the game factories to correctly give you the option to create a populated board by using the populated trait.  By default the game factory will produce an empty board.

This means that you will need to explicitly call the method when you want to populate the board.  When we create a new game for players, you also now have the option of designing a layout that first starts with an empty board, and then upon some event, populates.

As mentioned to Zach, it would look something like this:

```
game = Game.new
game.save
game.populate!
```

In a future PR I will resurrect FactoryGirl for all testing so that it will be much easier to use FactoryGirl's DSL to make future tests.


aleksgorbenko:
The way our front end view and back end are connected right now - we can't change the sequence of piece creation. The reason for that is we have a formula in #40 that converts IDs in the front end to the IDs in the DB and the other way around. If we change the sequence now - the front end view will be totally scrambled.

At the moment the IDs are assigned like this:

1  -   8
8  -  16
empty cells
16 - 24
24 - 32

The same ID has each piece in the DB.

Based on how it is in the #40 your code could be simply changed to:

```Ruby
def populate!
  populate_left_black_half!
  populate_right_black_half!
  populate_black_pawns!
  populate_white_pawns!
  populate_left_white_half!
  populate_right_white_half!
end

private

# methods as they are in PR 40
```

Please, let's do that after the #40 is merged.

Let me know if you have any questions in the way its setup - I can explain on a call.

aleksgorbenko:
Ok 👍  from me as we have discussed this fix on a call and @ZachBerman sends one as well from sunny and Internet-less Golden, Canada.

I will merge.

PR #53, opened 2016-08-04
Castling Logic

chadbaum:
Supports allowing a king to initiate a castling move by attempting to `move!` into the corresponding locations.  Every validation is in there except check logic so it is not complete.  A king cannot make a castling move if he starts the turn in check, and importantly, cannot castle if any square between the origin and destination would make him attackable.  That will be implemented later.

There are 3 tests so far for this functionality.  I will add more when I revamp FactoryGirl to make it easier to test black and white pieces.

This is dependent on #52 so please do not merge this first.


app/models/king.rb
```diff
@@ -14,4 +14,37 @@ def valid_move?(x, y)
   def radial_move?(x, y)
     x_distance(x) <= 1 && y_distance(y) <= 1
   end
+
+  # Returns true if the king and rook have not moved,
+  # the location is a valid castling target, and there is
+  # no obstruction along the way. DOES NOT INCLUDE CHECK LOGIC.
+  def clear_castling_move?(x, y)
```
lasley: IMO this name is ambiguous. Maybe instead `is_castle_target?` or `valid_castle_target?`

spec/models/king_spec.rb
```diff
@@ -94,4 +94,74 @@
...
+      castling_rook.update(moved: true)
+      expect(castling_king.move!(6, 7)).to eq false
+      expect(castling_king.x_position).to eq 4
+      expect(castling_king.y_position).to eq 7
+      expect(castling_king.moved).to eq false
+      expect(castling_rook.x_position).to eq 7
+      expect(castling_rook.y_position).to eq 7
+      expect(castling_rook.moved).to eq true
+    end
+  end
```
lasley: Looks to me like you have a different logic path for castling white/black, but I don't see any tests for black. Are we sure those conditionals work?

spec/models/king_spec.rb
```diff
@@ -94,4 +94,74 @@
...
+        type: 'Rook',
+        color: 'white',
+        x_position: 0,
+        y_position: 7
+      )
+    end
+
+    it 'should return true on valid castling' do
+      castling_rook.save
+      expect(castling_king.move!(6, 7)).to eq true
```
lasley: You should try to limit to one `expect` per `it`. Otherwise, one condition failing means that the rest of the suite will not run and you will not have a clear picture of what is actually broken.

spec/models/king_spec.rb
chadbaum: Yeah, I was naughty and didn't give full coverage because I was trying to do it all in one pass in a separate PR where I reinstate FactoryGirl for all tests.  However, I am ashamed and will add tests for black coverage.

app/models/king.rb
chadbaum: This conforms with my clear_ modifier that modified all the original moveset validations to show that it was also checking for obstruction.  So if we remove clear here, we should remove it everywhere.  Clear essentially means = unobstructed, as I've used it.  Technically a castling move HAS to be unobstructed so you could argue clear could be dropped here, but for the other helper methods it cannot be so similarly implied because they are helper methods and not directly implementing a known chess rule.

chadbaum:
@aleksgorbenko Please take a look at the last 3 remaining failed tests.  I spent hours trying to integrate your code with mine but 1 of your tests and 2 of my tests are failing due to a stack level too deep error related to `checked_cell`.  Can you take a look and see if you can find out what is going on?  One other thing that may help, when I disable the knight test it also solves one of my tests, so I would also make sure your variables aren't polluting a global namespace somehow.

chadbaum:
Fixed bugs with Aleks on call after merging latest checkmate branch.  Ready for review.  Added tests for black and white kings and queen-side and king-side castling attempts with obstruction and non-obstruction.

lasley:
LGTM 👍 

IMO you should use this PR to practice the art of squashing your commits before merge.

http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

Note you'll need a `git push -f` (force) once you do, because Git will say remote branch is ahead X commits and local is by 1 (that's all the commits you squashed remote, vs the one local, you want to blast the remote).

chadbaum:
@lasley Is squashing commits in this way the desired result for PRs so that each PR into master eventually gets represented by only one commit in the history?

lasley:
Yup, it creates an implicit change log

lasley:
Typically you want to submit a PR with one commit, then someone reviews and you drop another commit fixing whatever, rinse repeat that till thumbs, then squash them all before merge

chadbaum:
Good to know since up until now it's been wild west with regard to commit practices.  I will squash and re-up.

lasley:
👍

PR #56, opened 2016-08-05
Check state

aleksgorbenko:
General check for Check state of the game. The method is added in the Game model. All tests are passing and using the existing logic. The reason why tests have to have 2 kings is because every game would have two and the `check?` method creates 2 variables for the white and black king.

At the moment, Rubocop says the method has too many lines, so I consider splitting it into two `black_check?` and `white_check`. Tests will have to be edited too of course. Ideas on how to shorten it to 10 lines down from 14?


aleksgorbenko:
Ok, have split the method into 2 and fixed the tests - all is working.

Tests are checking for check for horizontal, vertical, diagonal and L-shape (Knight) moves.

app/models/game.rb
```diff
@@ -14,6 +14,30 @@ def populate!
     populate_right_white_half!
   end
 
+  def white_check?
```
lasley: Couldn't we combine these two methods into `check?`, then just accept an argument for the color?

app/models/game.rb
```diff
@@ -14,6 +14,30 @@ def populate!
...
+    white_king = pieces.find_by(type: 'King', color: 'white')
+    pieces.each do |piece|
```
lasley: Loop all the pieces? Couldn't we save some horsepower by first filtering on opposite color?

app/models/game.rb
aleksgorbenko: Rubocop is angry at this. I had it combined before.

Also, i thought we could use this logic in games_controller by adding

```Ruby
def show
  if white_check?
    flash[:white_check] = "message of check that we can show in the view"
```

app/models/game.rb
```diff
@@ -14,6 +14,30 @@ def populate!
...
+      if piece.enemy?(white_king) &&
+         piece.valid_move?(white_king.x_position,
```
lasley: This looks like an infinite recursion with `valid_move?`, as both `black_check?` and `white_check?` are called in `valid_move?`.

lasley: Edge case that would generate this would be two head-to-head kings. That's kind of a stupid situation, but so are users. Please also add a test for what happens if a king tries to move into a check position with a king

spec/models/game_spec.rb
```diff
@@ -20,4 +20,135 @@
...
+        type: 'Knight',
+        color: 'white',
+        x_position: 6,
+        y_position: 7,
+        moved: false
+      )
+    end
+
+    it 'should return true if King is under check horizontally' do
+      expect(queen.move!(1, 4)).to eq true
```
lasley: You should try to limit to one `expect` per `it`. Otherwise, one condition failing means that the rest of the suite will not run and you will not have a clear picture of what is actually broken.

app/models/game.rb
aleksgorbenko: Ops, you are right - will do.

app/models/game.rb
aleksgorbenko: This logic is in #57 - prevent any king to step into any cell that is in the line of attack of opposite color piece, including the King.

app/models/game.rb
lasley: Interesting, I'm not sure if I understand what Rubocop was complaining about. Adding an argument does nothing to the lines of code or the complexity of the method. Are you sure you didn't refactor beyond this?

aleksgorbenko: I was getting method is too long (11/10) lines. Didn't see the way to make it work in one, so split it into two.

lasley: Yeah but what I'm saying is that you can add an argument without adding any lines, so I feel that this method was refactored beyond just removing an argument for a static string.

aleksgorbenko:
Ok made most of the fixes as per comments above. For now reverted back to the original Kings valid_move? method - will fix the logic in #57 to prevent king into moving into the check.

app/models/game.rb
chadbaum: LoL aleks this is what i suggested before we hung up! :0

aleksgorbenko: This is what I had initially:

```Ruby
  def check?(color)
    black_king = pieces.find_by(type: 'King', color: 'black')
    white_king = pieces.find_by(type: 'King', color: 'white')
    if color == 'white'
      pieces.select { |p| p.color = 'white' }.each do |piece|
        return true if piece.valid_move?(black_king.x_position,
                                         black_king.y_position)
      end
    elsif color == 'black'
      pieces.select { |p| p.color = 'white' }.each do |piece|
        return true if piece.valid_move?(white_king.x_position,
                                         white_king.y_position)
      end
    end
    false
  end
```
It is 14 lines and is too long for Rubocop.

Then I have removed the top 2 variables outside, so had 12 lines in total. Beyond that...I am not sure how to make it shorter.

@chadbaum with the usage of valid_move? - yes, I remember we arrived at it together, when you saw the older method that I had.

chadbaum: nah i just meant passing in color as a parameter

chadbaum: look at the parts of your code that are repetitive and think of ways to refactor using helper methods to resolve the problem of method being too long for rubocop

aleksgorbenko: Ok, I managed to fit it in line limit but I still have an ABC error in rubocop [15.3/15]. It is only *0.3* points away from being Rubocop compliant.

```Ruby
def check?(color)
    if color == 'white'
      blk_pieces.each do |p|
        return true if p.valid_move?(w_k.x_position, w_k.y_position)
      end
    else
      wht_pieces.each do |p|
        return true if p.valid_move?(b_k.x_position, b_k.y_position)
      end
    end
    false
  end

  private

  def b_k
    pieces.find_by(type: 'King', color: 'black')
  end

  def w_k
    pieces.find_by(type: 'King', color: 'white')
  end

  def wht_pieces
    pieces.select { |p| p.color = 'white' }
  end

  def blk_pieces
    pieces.select { |p| p.color = 'black' }
  end
```

Here is the reference for ABC: http://c2.com/cgi/wiki?AbcMetric

Firstly, I know having `b_k` and `w_k` variables for Kings is not great, but that was the only way I could fit code in one line.

Secondly, I agree, having a color param is better (it's just I couldn't fit the whole method in before). Now, I can actually use this logic to prevent each Piece to open a check to their own king with this in our `move!` method:

`return false if x_position = x && y_position = y && check?(color)`

I need to help to refactor this code to make it work.

I got the checkmate logic all written up too (no PR yet), just need to finish this up and then carry on with checkmate and tests for it.

aleksgorbenko: Ignore my previous comment - I made it work in a much more elegant and Rubcop compliant way :)

lasley:
Looks like we have conflicts

aleksgorbenko:
All fixed - ready to merge

chadbaum:
:+1:

Se7enB2st:
👍

PR #57, opened 2016-08-06
Prevent king move into check

aleksgorbenko:
Based off #56 

The method that takes x,y coordinates, chooses all pieces that are the opposite color of the given King and iterates through each piece to see if it can move to the x,y coords. If true - King would be in check and move is invalidated.

Tests are passing, had to fix one test as the new validation prevented King from moving there.

If all looks good - please first review and merge the #56 

I will be working on the checkmate logic most likely based on these two PRs as well.


aleksgorbenko:
Alright, fixed the valid_move? for the King, now it cannot enter the cell that is in the line of attack.
This PR contains the latest fixes from #56 
Added flash messages that can be then picked up from `show` controller in the front view

aleksgorbenko:
The logic to prevent the King from moving into the check is working and tested. 

I started to implement the logic to invalidate the move of any piece if it opens up its own King (this is what is currently failing)

chadbaum:
As discussed:

```
def check_state(x, y, color)
    old_x = x_position
    old_y = y_position
    update(x_position: x, y_position: y)
    result = game.check?(color)
    update(x_position: old_x, y_position: old_y)
    result
  end
```

aleksgorbenko:
Thanks @chadbaum once again - the method is now working and passes the tests.

This PR has the following logic built-in:

- King prevention from going to checked cell
- Any piece prevention from opening their own King

Note: the reason why the last test in this PR is so long, is because the check method is checking for king variables after every move and loops through all enemy pieces. Therefore, for the test to work properly we need both Kings and several other pieces.

I have also shortened one existing method to make the PR compliant with rubocop. We will probably need more of this in the future, but that's for another PR.

app/models/game.rb
```diff
@@ -17,13 +17,18 @@ def populate!
...
-    king = king(color)
+    king = pieces.find_by(type: 'King', color: color)
     enemy_pcs(color).each do |p|
       return true if p.valid_move?(king.x_position, king.y_position)
     end
     false
   end
 
+  # selects enemy pieces that are not captured
+  def enemy_pcs(color)
```
lasley: I recommend not using erroneous abbreviations in code - it makes it so that you have to mentally try to figure out what abbreviations (if any) you used for things. Not a blocker or anything, just a best practice.

app/controllers/games_controller.rb
```diff
@@ -11,7 +11,13 @@ def create
   end
 
   def show
+    @game = Game.find(params[:id])
     @piece_positions = Game.find(params[:id]).pieces
+    if @game.check?('black')
+      flash[:black_check] = 'Black King is under the Check!'
```
lasley: s/King is under the/King is in

app/controllers/games_controller.rb
```diff
@@ -11,7 +11,13 @@ def create
...
+    elsif @game.check?('white')
+      flash[:white_check] = 'White King is under the Check!'
```
lasley: s/King is under the/King is in

lasley:
Few comments, but this code looks awesome 👍

Se7enB2st:
👍

PR #59, opened 2016-08-06
User Dashboard, Index page re-format

ZachBerman:
-Generate players controller, hook up show actions and route
-Create user dashboard and stats
-Generate static_pages controller
-Relocate current index page to static_pages#index
-If user is logged in, redirect to games#index
-Further re-factored CSS adding devise..scss for devise views

****Will update stats further when supported, along with create/join functionality & styling in subsequent PR


chadbaum:
What is players controller for?  Would expect that stuff in users.

ZachBerman:
We don't have a users controller currently so I followed the wireframe as players#show

ZachBerman:
Switched over to users controller, also added a test for redirect for signed in users.

app/views/games/index.html.erb
```diff
@@ -1,182 +1,51 @@
 <div class="container-fluid">
-  <section id="register">
-    <div class= "row">
+  <section id="games-list">
+    <div class="row">
       <div class="col-sm-5 col-sm-offset-1">
```
lasley: Offsetting one (`col-sm-offset-1`) on the first column ? I think maybe this should be on the second col in order to provide spacing between the two?

app/views/games/index.html.erb
```diff
@@ -1,182 +1,51 @@
...
-        </div>
-        <br />
-        <div class="form-group">
-          <%= f.button :submit, "Create An Account", class: "btn btn-success form-control" %>
-        </div>
-      <% end %>
+      <div class="col-sm-5">
+        <h2 class="join-headline text-center">Your opponent awaits you! Join the battle!</h2>
+          <h4>Available game<a class="btn btn-success pull-right" href="#" role="button">Join game!</a></h4>
+          <br class="clr" />
```
lasley: Even though you see them in the Rails scaffolds all the time, `<br />` is bad practice because it is not [semantic](https://en.wikipedia.org/wiki/Semantic_Web). They're basically only valid when outputting data that explicitly needed a line break, but still it's better to use `<pre>` in those instances.

In order to guarantee that objects are in different visual blocks in all browsers, you should make sure they are in different logical blocks (aka utilize encapsulation to provide your visual looks, similar to the way bootstrap does - think Russian nesting doll).

In most cases, the direct replacement for a `<br/>` is `<p></p>`

app/views/games/index.html.erb
```diff
@@ -1,182 +1,51 @@
...
-          <%= f.password_field :password_confirmation, class: 'form-control', required: true, placeholder: 'Confirm Password' %>
-        </div>
-        <br />
-        <div class="form-group">
-          <%= f.button :submit, "Create An Account", class: "btn btn-success form-control" %>
-        </div>
-      <% end %>
+      <div class="col-sm-5">
+        <h2 class="join-headline text-center">Your opponent awaits you! Join the battle!</h2>
+          <h4>Available game<a class="btn btn-success pull-right" href="#" role="button">Join game!</a></h4>
```
lasley: This would look better & scale for mobile if you instead nest a row div, then put the available game in a col-6 w/ `text-left` class and the Join game button in another col-s w/ `text-right` class

app/views/games/index.html.erb
```diff
@@ -1,182 +1,51 @@
...
-      <div class="col-sm-2 col-sm-offset-1 col-xs-offset-2 col-xs-8">
-        <%= image_tag 'chad.png', class: 'img-responsive team-profile' %>
-        <h5 class="text-center">Chad</h5>
-          <div class="text-center">
-          <%= image_tag 'linked_in.png' %> &emsp;
-          <%= link_to 'https://github.com/chadbaum' do %>
-            <%= image_tag 'github.png' %>
-          <% end %>
-        </div>
+      <div class="stats col-sm-offset-2 col-sm-2">
```
lasley: Why deleting the `col-xs` stylings?

app/views/layouts/application.html.erb
```diff
@@ -1,17 +1,17 @@
 <!DOCTYPE html>
 <html>
+  <body class="site">
```
lasley: This was in the correct spot. Head is outside of body. http://www.w3schools.com/tags/tag_head.asp

app/views/users/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="container-fluid">
+  <section id="player-dashboard">
+    <div class="row">
+      <div class="user col-sm-offset-1 col-sm-2">
+        <%= image_tag 'user.png', class:'img-responsive' %>
+        <h4><%= current_user.email %></h4>
+        <h4>Member Since: <%= current_user.created_at.strftime("%B %d, %Y") %></h4>
+        <a class="btn-block btn btn-info account-settings" href="#" role="button">Account Settings</a>
+      </div>
+      <div class="boxed col-sm-8 col-xs-10 col-sm-offset-0 col-xs-offset-1">
```
lasley: offset-0 is not a thing http://getbootstrap.com/css/#grid

app/views/users/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <h2 class="text-center">Hello There!</h2>
+        <div class="col-sm-offset-1">
+        <h3><b>Your Statistics</b></h3>
+        <br />
```
lasley: Find a way to separate w/o `<br>`

app/views/users/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <h4>Games Played: <%= current_user.games.count %></h4> 
+        <h4>Wins</h4>
```
lasley: Too many header tags IMO

app/views/users/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <div class="user col-sm-offset-1 col-sm-2">
+        <%= image_tag 'user.png', class:'img-responsive' %>
+        <h4><%= current_user.email %></h4>
+        <h4>Member Since: <%= current_user.created_at.strftime("%B %d, %Y") %></h4>
+        <a class="btn-block btn btn-info account-settings" href="#" role="button">Account Settings</a>
+      </div>
+      <div class="boxed col-sm-8 col-xs-10 col-sm-offset-0 col-xs-offset-1">
+        <h2 class="text-center">Hello There!</h2>
+        <div class="col-sm-offset-1">
+        <h3><b>Your Statistics</b></h3>
```
lasley: `<strong>` instead of `<b>`

app/views/users/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <br />
+        <h4>Games Played: <%= current_user.games.count %></h4> 
+        <h4>Wins</h4>
+        <h4>Losses</h4>
+        <h4>Draws</h4>
+        <h3><b>Historical Battles</b></h3>
```
lasley: `<strong>` instead of `<b>`

app/views/layouts/application.html.erb
ZachBerman: I'm referencing this [https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/](url) to keep the footer pinned to the bottom...

app/views/users/show.html.erb
ZachBerman: From the documentation - "You can also override offsets from lower grid tiers with .col-*-offset-0 classes." - However I'm not sure I'm using it correctly? I will go over it regardless

app/views/layouts/application.html.erb
lasley: That has a `<header>` inside of the body, not `<head>`

app/views/users/show.html.erb
lasley: Ah yeah I forgot they added this on release, but I don't think it's being used right here - and I've honestly never seen it used in the wild.

I assume this is achieving the effect of centering the element while it's on an `xs` screen?

app/views/games/index.html.erb
ZachBerman: This whole thing will be re-styled in subsequent PR.

app/views/layouts/application.html.erb
ZachBerman: :+1:

app/views/games/index.html.erb
ZachBerman: I was trying to style having 5 pictures horizontally across xs screens I believe - didn't work out how i wanted.

lasley: The XS is on a different grid level than the SM. You would typically want pictures to be stacked on SM and XS grid levels, then maybe two wide on MD, then four on LG

lasley: http://www.zeninvader.com/css/bootstrap-3-grid-system-explained

ZachBerman: Yeah - I know - but this was changing on the xs level as well, any clue why?

app/views/games/index.html.erb
ZachBerman: In my next PR this is all re-done, looping through available games. This was pushed up in talking with Chad as I believe it was blocking him for something.

app/views/games/index.html.erb
ZachBerman: This is just to offset 1 small column from left side of page. Is this not a good practice? Find myself using it frequently to keep things off the edges...

lasley: I figured that was the case. The issue with this strategy is that it's only emulating a center and not actually providing it.

Does a `col-sm-6 text-center` not achieve the desired results?

ZachBerman: Not exactly, but I see what you are saying. It's a non-issue for the moment because we decided to get rid of the picture there to allow for more room to list our available games. This will be in the PR to follow.

lasley:
👍

chadbaum:
:+1:

PR #61, opened 2016-08-08
HTML/CSS Cleanup

chadbaum:
- Created html partials for navbar and footer, separating the code out from application.erb so it is now much more readable.
- Fixed bug where body was being declared before head
- Moved games controller related SCSS code from custom _game.scss file to Rails convention games.scss file
- Split SCSS code into several files:  Games for games controller related code, Navbar for navbar partial, Footer for footer partial, Global as a catch-all for styles we anticipate using with both games and static_pages controllers, and Mixins.
- Renamed from 'css.scss' to 'scss' due to deprecation warnings from RSpec.
- Fixed some clashing CSS code causing the chessboard to not display correctly (removed some media queries for now -- we can discuss where and how to add them back in later).  Chessboard, imo, should not be fully responsive.


aleksgorbenko:
All looks fine, the only thing is - I am not sure why you removed the media queries for the board? It does not make the board exactly responsive, but it makes it little bigger on larger screens and makes it fit on smaller devices. I was checking it my screen and frankly when I looked at the same version as it is on mobile on the desktop screen - it looks just unnaturally small.

And seems it needs the latest master re-merged once again to be good to go.

chadbaum:
Yeah, I removed the media queries to get things working at a base level, and then my idea was to add them back in in the later design sweep.

ZachBerman:
:+1:

aleksgorbenko:
Ok, as you wish. My media queries are offended by such mistreatment though.
👍

PR #62, opened 2016-08-11
Turn base logic

Se7enB2st:
Add:
First Turn:
At the start of the game, set turn number to 1 and turn to white.

Turn Base Logic:
Increment turn number at the end of turn.
Alternate color at the end of turn (black or white).

Integrate turn-base logic into move! 

It will return false if a moving piece's color doesn't match the game color. 
This prevents player moving on a wrong turn, moving in a continuous multiple turns, etc.

### DIFF
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -33,14 +33,20 @@ class Game < ActiveRecord::Base
     players.find_by(color: 'black')
   end
 
-  def end_turn!
+  # At the end of turn
+  # Alternate the color between white and black (first turn color is white)
+  # Increment turn number by 1
+  def end_turn!(color)
+    next_turn_color = color == 'white' ? 'black' : 'white'
+    update(color: next_turn_color)
     increment!(:turn)
   end
 
   private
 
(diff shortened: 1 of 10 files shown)


chadbaum:
As mentioned, no need for color property for game -- just assume `turn.even?` is black and `turn.odd?` is white.

chadbaum:
Can you explain what is in this PR vs the En Passant one?  You indicate that you already implemented turn-based logic in En Passant.  Is the other PR dependent on this one?  I think it would be easiest if you instead made 1 PR for turn-based logic, and then 1 PR for en passant based on that.

Also, neither PR is ready to be merged.

Se7enB2st:
Yes, En Passant PR is dependent on this PR. After This PR is done, I can merge and refractor the En Passant PR.

chadbaum:
So all turn-based logic is in this PR?  None is in the other PR?  If you want us to merge this one first, please note that the other one is dependent with the normal [] syntax in header.  

Also please remove the extra column as mentioned above.

PR #63, opened 2016-08-15
Checkmate

aleksgorbenko:
The most important change - I actually found the bug in the checked_cell? method which in rare cases will return `false` if the king is checked. Hence, I had to refactor and integrate check_state everywhere instead as it will be more accurate in all cases, including edge ones.

At this stage I have
- 1 test passing for checkmate (need to write more)
- ABC error on `checked_coords` method
- 101/100 lines on Piece.rb (due to the fact that I had to move things around)

I tried to move `check_state` method to the Game model, but unsuccessfully. Hence I had to move `checkmate?` method to Piece model as well.


aleksgorbenko:
Methods refactored. Rubocop is now passing, the test is working. Will write more checkmate tests now.

chadbaum:
I would expect check logic and checkmate logic to be in the game object since it's more properly a behavior of the game itself and not any individual piece -- what was the problem you had with inserting them into the game model?

aleksgorbenko:
The `check_state`, has calls on Piece properties like `x_position` and `update(x_position: X, y_position: Y)`. This method has to be called on a piece and not the game object. 

Checkmate logic uses this `check_state` method to determine if all cells around the king are in check. So I had to move it to Piece model, so that it can use the `check_state`.

chadbaum:
I believe at this point that we should refactor everything after we have the app working, which means focus on front end and integrating websockets.  However, afterwards, we can start refactor PR's (for those interested in continuing work on project) to tidy up.  

That said, I'm confident that this can still be moved to the game model without any issue with invoking piece logic behavior.

aleksgorbenko:
@chadbaum I agree with you. I do believe it can be moved and should. I am up for refactoring PR's later on after we got the app working and after the Web Sockets.

aleksgorbenko:
I cannot push anymore to this branch. I just refactored some methods, added comments and removed redundant method.

I get this:

```Ruby
error: failed to push some refs to 'git@github.com:TeamStoic/StoicChess.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
```

I pulled the latest master, but it doesnt seem to fix anything.

chadbaum:
It's because I merged master into the branch from github.  Just `git pull origin <this branch>` then push up again.

aleksgorbenko:
Ok - works now, thanks!

lasley:
This looks good to include in the release. I have a few comments, but I'm going to hold them because it's more important to get on master at this point 👍

chadbaum:
:+1:

PR #64, opened 2016-09-03
Createjoin game

ZachBerman:
-Add functionality to create new game & join existing game
-Add stats to games#index


app/controllers/games_controller.rb
```diff
@@ -1,13 +1,17 @@
...
+  before_action :authenticate_user!, only: [:create]
 
   def index
+    @games = Game.all
   end
 
   def create
-    @game = Game.new
+    @game = Game.create
+    @game.white.update(user_id: current_user.id)
```
lasley, line 12: ~~What is this doing? IIRC there isn't a `white` attribute on game~~ NVM

app/controllers/games_controller.rb
```diff
@@ -32,6 +32,12 @@ def update
     @piece.move!(x, y)
   end
 
+  def join
+    @game = Game.find(params[:id])
+    @game.black.update(user_id: current_user.id)
```
lasley: What is this doing? IIRC there isn't a `black` attribute on game

app/models/game.rb
```diff
@@ -40,6 +40,14 @@ def checkmate_coords(x, y)
     end
   end
 
+  def white
```
lasley, line 43: These helpers seem unnecessary and are ambiguously named

app/controllers/games_controller.rb
lasley, line 12: This is performing three db writes + a read, and is inefficient. The game should be created and populated with the player all in one fell swoop.

app/models/game.rb
ZachBerman: It was my understand that the way we have our model set up, these are necessary to populate one "player" on the board and all the color pieces they come with...

lasley: Oh I see, these were already defined. What was the change here then?

ZachBerman: I had to make the methods public so I could access them.

lasley:
Ahh cool works for me then 👍

